PR #7, opened 2021-01-05
Added refund bank balance on island disband

BomBardyGamer:
When you disband your island, the balance will now be refunded to the owner of the island. (https://trello.com/c/V5kzyJiA)

I was originally using `islandBank.withdrawMoney(getOwner(), islandBank.getBalance(), null)`, but the message "Bank: BomBardyGamer has withdrawn $100,000 from the island bank" or whatever it is felt a bit out of place.

Feel free to request changes if you want.

Edit: I have tested this with both `/is disband` and `/is admin disband` using Essentials + Vault for economy (though any economy that supports Vault should work of course).

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
@@ -1138,6 +1138,8 @@ public final class SIsland implements Island {
             }).forEach(superiorPlayer::resetMission);
         });
 
+        plugin.getProviders().depositMoney(getOwner(), islandBank.getBalance());
+
         plugin.getMissions().getAllMissions().forEach(this::resetMission);
 
         resetChunks(true);


BomBardyGamer:
I wasn't actually 100% sure if this was how you wanted to do it, but refunding the entire balance to the island owner made the most sense. Let me know if you had other ideas :)

OmerBenGera:
@BomBardyGamer You cannot use IslandBank#widthrawMoney for a couple of reasons:
1) You're not withdrawing money from the bank technically.
2) IslandBank#widthrawMoney makes a few database calls that can be avoided by using other methods.
3) The withdraw message is sent to the players of the island which doesn't make sense either.

You should use the economy hook to give money to a player without actually withdrawing it from the bank :)
After you make that change, lmk and I'll add a configurable message that alerts the owner that he received X amount of money and we can close this

PR #9, opened 2021-01-06
Added island leave confirm menu

BomBardyGamer:
When you attempt to leave your island, there is now the option for a configurable leave confirmation menu to appear, to confirm if you want to leave or not. (https://trello.com/c/O3vKuIRO)

Fully tested with both 1.8 and 1.13, with and without the menu, and it works fine.

Feel free to request changes if you want.


BomBardyGamer:
Sorry about the confusion, I realised in the old commit that git didn't properly stage my change to `previous-menu`, so I soft reset and then had to hard reset and copy over the files again

BomBardyGamer:
Also, if you're wondering why these may be in a weird order (not ordered in order of Trello cards), it's because I'm doing the ones I know how to first xD

PR #12, opened 2021-01-07
Added teleport warmup bypass permission

BomBardyGamer:
When you either warp (/is warp) or teleport (/is home) to an island, there is now a permission (`superior.admin.bypass.warmup`) to bypass the warmup sequence (https://trello.com/c/O9K2hD7t).

Fully tested on Paper 1.12.2 with and without permission, using LuckPerms + Vault for permissions.


OmerBenGera:
Check if having the permission `superior.admin.bypass` also bypasses the warmup or not.
If it does, then fix it so only `superior.admin.bypass.warmup` will bypass it

BomBardyGamer:
Yeah does, will fix, hold up

PR #20, opened 2021-01-10
Added check for SSBSWM into spawn check

Paul19988:
Adding in this check avoids the server shutting down when it can't find the world whilst SSBSWM is generating it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 
 # Ignore Gradle build output directory
 build
+/gradle.properties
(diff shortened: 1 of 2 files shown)


OmerBenGera:
I don't understand what it fixes. You just removed the check for the world if the SlimeWorldManager is installed without doing any additional checks.
Mind explaining the original issue?

Paul19988:
The original issue would come when SSB would check for the spawn world it would just shutdown the server however the world may still exist in SWM it just hasn't loaded yet

OmerBenGera:
> 
> 
> The original issue would come when SSB would check for the spawn world it would just shutdown the server however the world may still exist in SWM it just hasn't loaded yet

But wait, performance wise, isn't it better to just have the spawn loaded all the time?

Paul19988:
Typically in most cases it would already be loaded other than on first boot where the world needs to be created

OmerBenGera:
We talked on Discord, will be implemented soon

PR #34, opened 2021-01-20
Fix for island member size having decimals

RemyyB:
Small fix for island members having decimals in the island top menu.


OmerBenGera:
I can't accept this PR.
The StringUtils#format adds commas and formats the numbers. You must use it.
Also, as you can see from the code, it already gets rid of numbers that end with .00

RemyyB:
I understand, but it doesn't get rid of the .00 as you can see in this screenshot (Paper 1.16.4 if that helps):

image

Thanks in advance ;)

OmerBenGera:
I can see what the issue is. The StringUtils#format assumes the number has a .00 at the end, but your number formatting sets it as ,00 at the end. Maybe there's way to know what the separator is by the number format?

OmerBenGera:
If you want to help me with the PR, I found this article: https://stackoverflow.com/questions/4713166/decimal-separator-in-numberformat
So instead of replacing with a .00, we need to use the decimal separator. I can do that later, but if you have the time, feel free to implement it yourself and post it here :)

RemyyB:
See the commit, hope this helps. Also I removed removed the grouping seperator if the amount of numbers are less or equal to 4 to prevent it being unclear in my language. You could remove it if you'd like ;)

OmerBenGera:
I pushed an update that addresses the issue. I didn't use your PR as I wanted to make things cached and work with the regex :)
Thanks for helping, appreciated :)

PR #279, opened 2021-05-27
Added ability to teleport straight to warp

ItchyDitchy:


OmerBenGera (changes requested):
Instead of copying code from MenuWarps, make a static method called "simulateClick" that takes the player, the warp and all the other information that is needed and make both files to use it.

src/main/java/com/bgsoftware/superiorskyblock/commands/CmdWarp.java
```diff
@@ -8,69 +8,91 @@
...
+		Pair<Island, SuperiorPlayer> arguments = args.length == 1 ? CommandArguments.getSenderIsland(plugin, sender) :
+			CommandArguments.getIsland(plugin, sender, args[1]);
+
+		Island island = arguments.getKey();
+		if(island == null)
+			return;
+
+		SuperiorPlayer superiorPlayer = plugin.getPlayers().getSuperiorPlayer(sender);
+		if (args.length == 3)
+			for (String warp : island.getIslandWarps().keySet()) {
```
OmerBenGera, line 71: You can use island.getWarp(String) to find a warp by it's name

src/main/java/com/bgsoftware/superiorskyblock/commands/CmdWarp.java
```diff
@@ -8,69 +8,91 @@
...
+	public void execute(SuperiorSkyblockPlugin plugin, CommandSender sender, String[] args) {
+		Pair<Island, SuperiorPlayer> arguments = args.length == 1 ? CommandArguments.getSenderIsland(plugin, sender) :
+			CommandArguments.getIsland(plugin, sender, args[1]);
+
+		Island island = arguments.getKey();
+		if(island == null)
+			return;
+
+		SuperiorPlayer superiorPlayer = plugin.getPlayers().getSuperiorPlayer(sender);
+		if (args.length == 3)
```
OmerBenGera, line 70: What if you run /is warp <warp-name>? It sounds like a valid command, but according to the code you must use /is warp <island> <warp-name>, which is not necessary.

OmerBenGera (changes requested):

src/main/resources/lang/en-US.yml
```diff
@@ -106,7 +106,6 @@ COMMAND_ARGUMENT_ENTITY: 'entity'
 COMMAND_ARGUMENT_ISLAND: 'island'
 COMMAND_ARGUMENT_ISLAND_NAME: 'island-name'
 COMMAND_ARGUMENT_ISLAND_ROLE: 'island-role'
-COMMAND_ARGUMENT_ISLAND_WARP 'island-warp'
```
OmerBenGera: Is there a reason why you removed this line?

OmerBenGera (changes requested):

ItchyDitchy:
> You removed all messages of the modules, why is that?
> Keep it them in the files please.

src/main/java/com/bgsoftware/superiorskyblock/commands/CmdWarp.java
ItchyDitchy: I'll go add a setting at config then I guess if they should prioritize the warp or the player.

src/main/java/com/bgsoftware/superiorskyblock/commands/CmdWarp.java
ItchyDitchy: Will implement...

ItchyDitchy:
.

src/main/resources/lang/en-US.yml
ItchyDitchy: Probably a mistake

OmerBenGera:
Any news regarding this?

PR #392, opened 2021-07-22
Add scale and rounding to SIslandBank#setBalance(BigDecimal, boolean)

WattMann:
IslandBanks have currently no precision set, that allows them to be more precise than will ever be necessary, when i was browsing SuperiorSkyblock database i noticed that there's a islandBank column with over 5k decimals, (see attached image). We can save up space by adding a 2 digit precision and round the decimals down.
image


OmerBenGera:
Please work with the dev build.

WattMann:
Okay

PR #399, opened 2021-07-26
Add rouding to SIslandBank

WattMann:
Currently, there is no precision limit, and that fills the database with useless decimals. I added 2 decimal precision and [HALF_DOWN](https://docs.oracle.com/javase/7/docs/api/java/math/RoundingMode.html#HALF_DOWN) rounding mode to SIslandBank#getBalance. See attached image, row with over 5000 decimals. 

image

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/island/bank/SIslandBank.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/island/bank/SIslandBank.java
@@ -25,6 +25,7 @@ import org.bukkit.entity.Player;
 
 import javax.annotation.Nullable;
 import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -50,7 +51,7 @@ public final class SIslandBank implements IslandBank {
 
     @Override
     public BigDecimal getBalance() {
-        return balance.get();
+        return balance.get().setScale(2, RoundingMode.HALF_DOWN);
     }
 
     @Override


src/main/java/com/bgsoftware/superiorskyblock/island/bank/SIslandBank.java
```diff
@@ -50,7 +51,7 @@ public SIslandBank(Island island){
 
     @Override
     public BigDecimal getBalance() {
-        return balance.get();
+        return balance.get().setScale(2, RoundingMode.HALF_DOWN);
```
OmerBenGera, line 54: The idea of rounding the balance is great, but I think there can be a better solution - you should round the balance when it's set, and therefore it will also get saved and displayed rounded.
This should be tested of course, but it should work.

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/island/bank/SIslandBank.java
martinprejsa: Originally that was my idea as well, but then i realized that if we want to round the setter as  well, we would need to add rounding to SIslandBank#increaseBalance too, since it acts as a private setter. So i figured it would be easier just to round the getter. Also there's no point of rounding the number twice (during get & set), getter rounding will do just fine. I can make the change if you really want to though.

OmerBenGera: Due to potential issues with threads and synchronizations, there is no way to combine the methods, and therefore yes - I prefer it to be in both setters and have a tiny of duplication code

martinprejsa: I shall do that then.

OmerBenGera: This can be reverted (there's no need for scale to getBalance anymore)

OmerBenGera:
You asked for a review but there's still need one request to be changed.

PR #410, opened 2021-07-31
Added ability to teleport straight to warp

Lukatjee:
Also had to modify the SkinsRestorerHook as it didn't want to compile with old code


src/main/java/com/bgsoftware/superiorskyblock/hooks/SkinsRestorerHook.java
```diff
@@ -4,13 +4,13 @@
 import com.bgsoftware.superiorskyblock.api.wrappers.SuperiorPlayer;
```
OmerBenGera, line 4: Do not edit files that has nothing to do with the feature itself.

src/main/java/com/bgsoftware/superiorskyblock/commands/CmdWarp.java
```diff
@@ -64,6 +68,22 @@ public void execute(SuperiorSkyblockPlugin plugin, CommandSender sender, String[
             return;
 
         SuperiorPlayer superiorPlayer = plugin.getPlayers().getSuperiorPlayer(sender);
+
+        if (args.length == 3) {
```
OmerBenGera, line 72: You assume the third parameter is always the warp's name, however the second parameter is optional.
What will happen if player will run /is warp <warp-name>? The logic you wrote will not occur, and if I am not mistaken, it should. The correct flow should be:
If there are 3 arguments -> third one is the warp's name
If there are two arguments -> check if the second argument is the warp's name.
If it's not the warp's name, then it's probably an island/player's name, and the old logic should occur regularly.
If it is a warp's name, then the player will be teleported to that warp.

Also, not all players can teleport to all warps, and there must be a check for that.
In the original code, the check is done in the requestObjects of MenuWarps.

src/main/java/com/bgsoftware/superiorskyblock/commands/CmdWarp.java
```diff
@@ -64,6 +68,22 @@ public void execute(SuperiorSkyblockPlugin plugin, CommandSender sender, String[
...
+
+            for (String warpName : island.getIslandWarps().keySet()) {
```
OmerBenGera, line 74: You should use Island#getWarp instead of looping through all the warps.

src/main/java/com/bgsoftware/superiorskyblock/menu/MenuWarps.java
```diff
@@ -46,21 +47,30 @@ public void onPlayerClick(InventoryClickEvent event, IslandWarp islandWarp) {
...
-                    return;
-                }
-
-                plugin.getProviders().withdrawMoney(superiorPlayer, plugin.getSettings().chargeOnWarp);
-            }
 
+            simulateClick(superiorPlayer, warpCategory.getIsland(), islandWarp.getName());
             Executor.sync(() -> {
                 previousMove = false;
-                superiorPlayer.runIfOnline(Player::closeInventory);
```
OmerBenGera, line 60: You removed this part and moved it to the simulateClick, however there's no guarantee now that previousMove=false will be called before closing the inventory, which might lead to the previous menu opening for the player.

OmerBenGera (changes requested):

OmerBenGera:
Also, please work with the dev branch and not master.

Lukatjee:
will be switching to dev branch

PR #421, opened 2021-08-05
improv. fix little issue with blocked commands

nivcoo:
Hello, i've fixed the issue if player send : /Jump or /JUMP if we set /jump in blocked visitors cmd

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/listeners/PlayersListener.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/listeners/PlayersListener.java
@@ -509,7 +509,7 @@ public final class PlayersListener implements Listener {
         Island island = plugin.getGrid().getIslandAt(e.getPlayer().getLocation());
 
         if(island != null && !island.isSpawn() && island.isVisitor(superiorPlayer, false) &&
-                plugin.getSettings().blockedVisitorsCommands.stream().anyMatch(cmd -> e.getMessage().contains(cmd))){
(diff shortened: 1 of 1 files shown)


src/main/java/com/bgsoftware/superiorskyblock/listeners/PlayersListener.java
```diff
@@ -509,7 +509,7 @@ public void onPlayerCommand(PlayerCommandPreprocessEvent e){
         Island island = plugin.getGrid().getIslandAt(e.getPlayer().getLocation());
 
         if(island != null && !island.isSpawn() && island.isVisitor(superiorPlayer, false) &&
-                plugin.getSettings().blockedVisitorsCommands.stream().anyMatch(cmd -> e.getMessage().contains(cmd))){
+                plugin.getSettings().blockedVisitorsCommands.stream().anyMatch(cmd -> e.getMessage().toLowerCase().contains(cmd))){
```
OmerBenGera: This can be optimized by caching the result of e.getMessage().toLowerCase() in a separate variable, and then check if this variable contains the cmd in it, instead of running e.getMessage().toLowerCase() every check.

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/listeners/PlayersListener.java
nivcoo: It's right, i will do that sunday

nivcoo:
is it fine like that ?

OmerBenGera (approved):

PR #493, opened 2021-08-20
improv. little error on return

nivcoo:
Hello, you can compare the older file you have inversed the role of island settings with this method compared to before.


nivcoo:
Also how works paper feature ? Because with that my spawner doesn't spawn any mobs, it's an issue or an option in paper ? Thx

PR #526, opened 2021-08-22
improv. return true if it's the spawn

nivcoo:
Hello, i forgot to change the return if the island is spawn


OmerBenGera:
It should be false if it's spawn and the spawn protection is disabled - as there's no protection to the spawn, including protection from mobs.

OmerBenGera:
There is another issue with detection of blocking entity spawning, and I am fixing it rn

Issue #76, opened 2021-02-17
Could not pass event ProjectileLaunchEvent to SuperiorSkyblock2 v1.5.2


Puruuuu:
https://pastebin.com/yRGn9hH9

OmerBenGera:
Hey, I pushed a potential fix - let me know if it fixes the issue.

PR #757, opened 2021-11-28
Attempt to disable logging if Console is making the transaction

Superalex0102:
Hey **Ome_R** again!

Referring to [#76](https://github.com/BG-Software-LLC/WildChests/pull/76), I tried to fix this issue, but I realized that there is a debugging method aswell that debugs to .session_logs file. 
I don't want to make any sudden changes, so I'm asking you how would you like to make a solution for this problem.

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/island/bank/SIslandBank.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/island/bank/SIslandBank.java
@@ -116,16 +116,19 @@ public final class SIslandBank implements IslandBank {
         SuperiorSkyblockPlugin.debug("Action: Deposit Money, Island: " + island.getOwner().getName() + ", Player: " + commandSender.getName() + ", Money: " + amount);
 
         UUID senderUUID = commandSender instanceof Player ? ((Player) commandSender).getUniqueId() : null;
+        boolean isConsole = !(commandSender instanceof Player);
 
         int position = transactions.readAndGet(SortedSet::size) + 1;
 
         BankTransaction bankTransaction = new SBankTransaction(senderUUID, BankAction.DEPOSIT_COMPLETED, position, System.currentTimeMillis(), "", amount);
         increaseBalance(amount);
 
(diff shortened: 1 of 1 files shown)


Superalex0102:
image
After running it for a few minutes with only 1 sell chest :(

src/main/java/com/bgsoftware/superiorskyblock/island/bank/SIslandBank.java
```diff
@@ -116,16 +116,19 @@ public BankTransaction depositAdminMoney(CommandSender commandSender, BigDecimal
         SuperiorSkyblockPlugin.debug("Action: Deposit Money, Island: " + island.getOwner().getName() + ", Player: " + commandSender.getName() + ", Money: " + amount);
 
         UUID senderUUID = commandSender instanceof Player ? ((Player) commandSender).getUniqueId() : null;
+        boolean isConsole = !(commandSender instanceof Player);
```
OmerBenGera, line 120: A better solution would be to make another method that takes as a parameter whether to save the transaction or not, instead of not saving it when console makes the payment. Also, addTransaction takes a similar parameter to whether to save the transaction to database or not, and it should be used instead.

OmerBenGera (changes requested):

OmerBenGera:
> image After running it for a few minutes with only 1 sell chest :(

This cannot be removed. You should reduce the amount of calls to the deposit method.

src/main/java/com/bgsoftware/superiorskyblock/island/bank/SIslandBank.java
Superalex0102: So I should make the same method but with an extra parameter?

OmerBenGera: > 

Yes, and make the original method to call the new one with a default parameter.

OmerBenGera:
Closing this as session logs file is not a thing anymore.
Thanks for contributing!

PR #1416, opened 2022-10-17
Update check boolean

Sulphinx:
Make a feature where you are allowed, to toggle the message about a new update.
Hence some of your versions, aren't compatible with our server versions.
Something alike:

#This boolean decides, whether or not to get the message on join, about a new update.
Update-check: true/false


OmerBenGera:
Not sure why you opened a PR from dev to master...
Anyways, server owners should stay up to date, therefore I am not removing the update norification.

Issue #1450, opened 2022-11-04
Mission Save Error


Pledix:
### Minecraft's Version

git-Pufferfish-22 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT)

### Plugin's Version

2022.9

### Describe the bug

`[00:51:18] [Craft Scheduler Thread - 122294 - SuperiorSkyblock2/ERROR]: [SuperiorSkyblock2] An unexpected error while saving mission data for farmer_5:
[00:51:18] [Craft Scheduler Thread - 122294 - SuperiorSkyblock2/WARN]: java.util.ConcurrentModificationException
[00:51:18] [Craft Scheduler Thread - 122294 - SuperiorSkyblock2/WARN]: 	at java.base/java.util.HashMap$HashIterator.nextNode(HashMap.java:1597)
[00:51:18] [Craft Scheduler Thread - 122294 - SuperiorSkyblock2/WARN]: 	at java.base/java.util.HashMap$EntryIterator.next(HashMap.java:1630)
[00:51:18] [Craft Scheduler Thread - 122294 - SuperiorSkyblock2/WARN]: 	at java.base/java.util.HashMap$EntryIterator.next(HashMap.java:1628)
[00:51:18] [Craft Scheduler Thread - 122294 - SuperiorSkyblock2/WARN]: 	at com.bgsoftware.superiorskyblock.missions.FarmingMissions.saveProgress(FarmingMissions.java:147)
[00:51:18] [Craft Scheduler Thread - 122294 - SuperiorSkyblock2/WARN]: 	at SuperiorSkyblock2-2022.9.jar//com.bgsoftware.superiorskyblock.mission.MissionsManagerImpl.saveMissionsData(MissionsManagerImpl.java:340)
[00:51:18] [Craft Scheduler Thread - 122294 - SuperiorSkyblock2/WARN]: 	at org.bukkit.craftbukkit.v1_17_R1.scheduler.CraftTask.run(CraftTask.java:101)
[00:51:18] [Craft Scheduler Thread - 122294 - SuperiorSkyblock2/WARN]: 	at org.bukkit.craftbukkit.v1_17_R1.scheduler.CraftAsyncTask.run(CraftAsyncTask.java:57)
[...]

PR #1507, opened 2022-12-13
fix(mission) try to fix ConcurrentModificationException !

NewZAZ:
Add iterator for fix ConcurrentModificationException

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/mission/MissionsManagerImpl.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/mission/MissionsManagerImpl.java
@@ -32,10 +32,7 @@ import javax.script.ScriptException;
 import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.Constructor;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
 import java.util.function.Consumer;
 
 public class MissionsManagerImpl extends Manager implements MissionsManager {
@@ -330,7 +327,7 @@ public class MissionsManagerImpl extends Manager implements MissionsManager {
 
         if (!dataFolder.exists())
             dataFolder.mkdirs();
-        for(Iterator<Mission<?>> missions = getAllMissions(); missions.hasNext(); ){
+        for(Iterator<Mission<?>> missions = getAllMissions().iterator(); missions.hasNext(); ){
             Mission<?> mission = missions.next();
             YamlConfiguration data = new YamlConfiguration();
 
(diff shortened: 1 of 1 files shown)


NewZAZ:
[#](https://github.com/BG-Software-LLC/SuperiorSkyblock2/issues/1450)

OmerBenGera:
The error happens because saveMissions is done async while saving data is done sync.
I dont think changing this to an iterator will have any effect on the error.

OmerBenGera:
The error happens because saveMissions is done async while saving data is done sync.
I dont think changing this to an iterator will have any effect on the error.

NewZAZ:
https://stackoverflow.com/questions/5847939/is-list-iterator-thread-safe

Iterator is thread safe and cannot get ConcurrentModificationException

OmerBenGera:
> https://stackoverflow.com/questions/5847939/is-list-iterator-thread-safe
> 
> Iterator is thread safe and cannot get ConcurrentModificationException

The concurrent modification is not because of this iteration.
The issue is that the missions modify the data while its being saved, causing the error to occur (from what Ive seen at least)
This is a fault of the way I designed the missions. I will need somehow to fix it without breaking the missions.
Still thinking about a way to do it.

NewZAZ:
Ok my bad.

iamnoksio:
> > https://stackoverflow.com/questions/5847939/is-list-iterator-thread-safe
> > Iterator is thread safe and cannot get ConcurrentModificationException
> 
> The concurrent modification is not because of this iteration. The issue is that the missions modify the data while its being saved, causing the error to occur (from what Ive seen at least) This is a fault of the way I designed the missions. I will need somehow to fix it without breaking the missions. Still thinking about a way to do it.

You literally just need to put a ConcurrentHashMap LMAO

OmerBenGera:
> > > https://stackoverflow.com/questions/5847939/is-list-iterator-thread-safe
> > > Iterator is thread safe and cannot get ConcurrentModificationException
> > 
> > 
> > The concurrent modification is not because of this iteration. The issue is that the missions modify the data while its being saved, causing the error to occur (from what Ive seen at least) This is a fault of the way I designed the missions. I will need somehow to fix it without breaking the missions. Still thinking about a way to do it.
> 
> You literally just need to put a ConcurrentHashMap LMAO
> 
> EDIT: getAllMission > missionMap = ConcurrentHashMap

As I said, this will not fix the issue.
The missions map is not changed during a session, and the error comes from modifing values in mission data holders.
When saving tracked data, the mission jars iterates through the data and save it. However, if in that time they also try to track data, it will throw the error.
Setting the map to a ConcurrentHashMap most of the time fix the issue, however thats not the case here.

iamnoksio:
you just dont know how to dev correctly dude

PR #1826, opened 2023-08-03
Update hook for UltimateStacker

ceze88:


OmerBenGera:
Thanks for the contribution, appreciated!

PR #1910, opened 2023-10-31
Update zh-CN.yml

SnowCutieOwO:
Re-translated the whole language file.
Now these texts will be more smooth to understand.


OmerBenGera:
A) This should be merged into dev branch
2) What is the reason for re-translating everything?

SnowCutieOwO:
> A) This should be merged into dev branch
> 2) What is the reason for re-translating everything?

1. Where is it? Please give me a link, thanks
2. The old one is outdated, and if I only add those untranslated, some of words may cause difference beacuse translators are not same, then if not, users' experience will become not so good.
Oh and another, there is multiple version of translating, put the translating here will make it "Official" so users needn't spend time to find one really match the latest version.

Thanks for understaing!

Issue #2008, opened 2024-02-25
All players currently on their island must sleep at the same time


jfoucher:
### Minecraft's Version

paper 1.20.4

### Plugin's Version

SuperiorSkyblock2-2023.3-b144.jar

### Describe the bug

When two  or more players are on their own islands and night comes, they have to all sleep at the same time, where I expected that each island should have independent sleep mechanism.

### To Reproduce

Create two (or more) islands. Have a player on each island. Wait for night to come. Try to sleep, the actionbar will show asleep 1/2 or whatever.
It should just let me sleep if I'm the only player on my island, and not depend on what goes on in other islands.

jfoucher:
Should be fixed by https://github.com/BG-Software-LLC/SuperiorSkyblock2/pull/2009

PR #2009, opened 2024-02-25
Do not force all players on different islands to sleep at the same ti…

jfoucher:
Do not force all players on different islands to sleep at the same time. Fixes https://github.com/BG-Software-LLC/SuperiorSkyblock2/issues/2008

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/listener/PlayersListener.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/listener/PlayersListener.java
@@ -40,21 +40,12 @@ import org.bukkit.event.block.Action;
 import org.bukkit.event.entity.EntityDamageByEntityEvent;
 import org.bukkit.event.entity.EntityDamageEvent;
 import org.bukkit.event.inventory.InventoryClickEvent;
-import org.bukkit.event.player.AsyncPlayerChatEvent;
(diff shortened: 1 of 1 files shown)


jfoucher:
@OmerBenGera Let me know if you'd like me to do something differently, or maybe add a config option for this.

OmerBenGera:
This will force daylight to the entire world, not only to the sleeping island.
This is not the best solution tbh, and in general making such thing is not gonna be easy.

Why not making a module to do that instead?

jfoucher:
Yeah you're right that this is a problem, but I could not find another way to make it work. Since the time is by world, each island should be it's won world for one to be at night and the other day, so indeed I don't think there is a proper solution to this problem. I use this PR on my server and, well for some players it will become day suddenly if players on another island went to bed,  but it is a compromise they are willing to accept.

OmerBenGera:
> Yeah you're right that this is a problem, but I could not find another way to make it work. Since the time is by world, each island should be it's won world for one to be at night and the other day, so indeed I don't think there is a proper solution to this problem. I use this PR on my server and, well for some players it will become day suddenly if players on another island went to bed, but it is a compromise they are willing to accept. I did not know it was possible to make modules for SuperiorSkyblock, I'll look into it.

Because some may don't want it, I prefer if you had it as a separate module.
I am closing this for now, thanks however for the contribution!

PR #2089, opened 2024-05-30
Add metadata method to Island classes and update build settings

MrLarkyy:
Added a new 'metadata' method to the Island, SIsland, DelegateIsland, and SpawnIsland classes, which returns a HashMap. Also, the build.gradle and settings.gradle were updated: Several inclusions were removed from settings.gradle and the version was updated in build.gradle.


MrLarkyy:
> Seems similar to the PersistenceDataContainer API

Yes, sorry it was not meant to be a PR to the official repo, but to my fork. However Ive been adding metadata to add non-persistent data. Ive been basically adding my custom handlers & other stuff and each island has had its own instance of those classes, so thats why Ive added metadata field. This helped me to track the handlers and I could have easily avoided making cache of UUID of islands with the custom class instances as I could have saved the instances directly to islands.

PR #2159, opened 2024-07-21
Add MiniMessage Support

Maxlego08:
Added support for [minimessage](https://docs.advntr.dev/minimessage/index.html) for Raw messages, this allows more configuration possibilities for messages.

A method to transform minecraft color codes into minimessage


OmerBenGera:
I made some changes to the way messages are constructed.
They are no longer a string, but of type MessageContent - which breaks the original messages into parts and can construct the message again without using replace methods (which may be heavy performance-wise).
We can think about a way that the MessageContent will parse MiniMessages somehow, but this PR will not work with that. Therefore, I am closing it.

Thanks for the contribution tho!

PR #2301, opened 2024-12-19
Add missing Bamboo & Cherry variants, lectern, decorated pot, copper doors, -trapdoors & crafter to interactables.yml

JorisvD01:
I noticed some of the new bamboo & cherry wood variants were missing from the interactables.yml (for example buttons, doors, trapdoors), as well as a lectern and enderpearl, so I just added those :)


src/main/resources/interactables.yml
```diff
@@ -95,6 +107,7 @@ interactables:
   - 'ENCHANTING_TABLE'
   - 'ENCHANTMENT_TABLE'
   - 'ENDER_CHEST'
+  - 'ENDER_PEARL'
```
OmerBenGera: I don't think this has any effect as only blocks are checked against the interactables list.
Other than that, looks good

OmerBenGera (changes requested):

JorisvD01:
Removed the enderpearl as it indeed didn't make a difference (just checked, sorry for that)

Also while i'm at it ill add all 1.21 copper doors / trapdoors

PR #2484, opened 2025-05-17
Update vi-VN.yml

maiminhdung:
Translate some keywords into Vietnamese.


OmerBenGera:
Thanks for the contribution!
I updated the lang files in recent commits, want to take a look at the new message file before I accept this MR?

PR #2496, opened 2025-05-24
Fix issue with custom schematics producing a NPE

Hephaisto-dev:
When using the API Schematic class, NPE was thrown due to null affected chunks passed to the database update function.
This PR fixes this issue with an empty list as a workaround and updates gradle versions as well as paperweightdevbundle versions


OmerBenGera:
The PR should be to dev.
Also, you changed 55 files for a simple fix.

Hephaisto-dev:
True i also upgraded version

OmerBenGera:
I pushed a fix similar to the NPE that this CR tries to fix.
I can't merge this as it has many changes not related to the fix itself.
If the issue still persists, open a bug report.
Thanks for the contribution!

PR #2500, opened 2025-05-31
Added option to disable island rating change

Lemvs:


src/main/java/com/bgsoftware/superiorskyblock/commands/player/CmdRate.java
```diff
@@ -75,6 +76,11 @@ public void execute(SuperiorSkyblockPlugin plugin, CommandSender sender, String[
             return;
         }
 
+        if (!plugin.getSettings().isChangeIslandRating() && island.getRating(superiorPlayer) != Rating.UNKNOWN) {
+            Message.RATE_ALREADY_GIVEN.send(superiorPlayer);
+            return;
+        }
+
```
OmerBenGera, line 83: Good one.
You also want to add this filter to the tabComplete method so it doesn't show islands that you cannot rate.

Lemvs:
Oh, okay. I'll try to do it 😁

Lemvs:
Sorry, but I messed up a lot here, I'll just do it again.

PR #2515, opened 2025-06-20
Dev1

Lemvs:
Added option to disable island rating change


Lemvs:
I did it wrong again, sorry >.<

PR #2530, opened 2025-06-28
Added transfer-confirm option

Lemvs:


Lemvs:
I made this according to confirm-ban/confirm-kick menu and everything works fine, so I don't know where this problem comes from

OmerBenGera:
> I made this according to confirm-ban/confirm-kick menu and everything works fine, so I don't know where this problem comes from

What problem?

PR #2536, opened 2025-07-02
Changed clear inventory on disband and join options

Lemvs:
I changed the way the options that cleared the inventory during disband/join work a bit. I separated them into the Ender Chest and inventory clearing options, which meant I had to change the names of these options to make them fit together, but I don't know if I can do that, so you can let me know if I did it correctly.

### DIFF
--- a/API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
+++ b/API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
@@ -237,12 +237,6 @@ public interface SettingsManager {
     String getStackedBlocksProvider();
 
     /**
-     * Whether inventory of island members should be cleared when their island is disbanded or not.
-     * Config-path: disband-inventory-clear
-     */
-    boolean isDisbandInventoryClear();
-
-    /**
      * All settings related to island-names.
      * Config path: island-names
      */
@@ -261,10 +255,28 @@ public interface SettingsManager {
     boolean isTeleportOnKick();
 
     /**
-     * Whether to clear players' inventories when they join a new island or not.
-     * Config-path: clear-on-join
+     * Whether to clear members Ender Chest when their island is disbanded or not.
+     * Config-path: clear-ender-chest-on-disband
      */
-    boolean isClearOnJoin();
+    boolean isClearEnderChestOnDisband();
+
+    /**
+     * Whether to clear players Ender Chest when they join a new island or not.
+     * Config-path: clear-ender-chest-on-join
+     */
+    boolean isClearEnderChestOnJoin();
+
+    /**
+     * Whether to clear players inventory when their island is disbanded or not.
+     * Config-path: clear-inventory-on-disband
+     */
+    boolean isClearInventoryOnDisband();
+
+    /**
+     * Whether to clear players inventory when they join a new island or not.
+     * Config-path: clear-inventory-on-join
+     */
+    boolean isClearInventoryOnJoin();
 
     /**
      * Whether players can rate their own island or not.
--- a/NMS/v1_12_R1/src/main/java/com/bgsoftware/superiorskyblock/nms/v1_12_R1/NMSPlayersImpl.java
+++ b/NMS/v1_12_R1/src/main/java/com/bgsoftware/superiorskyblock/nms/v1_12_R1/NMSPlayersImpl.java
@@ -41,11 +41,26 @@ public class NMSPlayersImpl implements NMSPlayers {
     }
 
     @Override
+    public void clearEnderChest(OfflinePlayer offlinePlayer) {
+        clearInventoryInternal(offlinePlayer, false, true);
+    }
+
+    @Override
     public void clearInventory(OfflinePlayer offlinePlayer) {
+        clearInventoryInternal(offlinePlayer, true, false);
+    }
+
+    public void clearInventoryInternal(OfflinePlayer offlinePlayer, boolean inventory, boolean enderchest) {
         if (offlinePlayer.isOnline() || offlinePlayer instanceof Player) {
             Player player = offlinePlayer instanceof Player ? (Player) offlinePlayer : offlinePlayer.getPlayer();
-            player.getInventory().clear();
-            player.getEnderChest().clear();
+
+            if (inventory) {
+                player.getInventory().clear();
+            }
+            if (enderchest) {
+                player.getEnderChest().clear();
+            }
+
             return;
         }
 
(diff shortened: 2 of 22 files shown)


API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
```diff
@@ -236,12 +236,6 @@ public interface SettingsManager {
      */
     String getStackedBlocksProvider();
 
-    /**
-     * Whether inventory of island members should be cleared when their island is disbanded or not.
-     * Config-path: disband-inventory-clear
-     */
-    boolean isDisbandInventoryClear();
-
```
OmerBenGera, line 244: Do not delete the method.
Instead mark it as deprecated, change the comment to reflect what it really returns (for example, return true if one of the options are true) and in the future it will be deleted

API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
```diff
@@ -261,10 +255,28 @@ public interface SettingsManager {
...
+     * Whether to clear players inventory when their island is disbanded or not.
+     * Config-path: clear-inventory-on-disband
+     */
+    boolean isClearInventoryOnDisband();
+
+    /**
+     * Whether to clear players inventory when they join a new island or not.
+     * Config-path: clear-inventory-on-join
      */
-    boolean isClearOnJoin();
```
OmerBenGera, line 267: Do not delete the method.
Instead mark it as deprecated, change the comment to reflect what it really returns (for example, return true if one of the options are true) and in the future it will be deleted

NMS/v1_12_R1/src/main/java/com/bgsoftware/superiorskyblock/nms/v1_12_R1/NMSPlayersImpl.java
```diff
@@ -40,12 +40,27 @@ public NMSPlayersImpl(SuperiorSkyblockPlugin plugin) {
...
+    public void clearEnderChest(OfflinePlayer offlinePlayer) {
+        clearInventoryInternal(offlinePlayer, false, true);
+    }
+
     @Override
     public void clearInventory(OfflinePlayer offlinePlayer) {
+        clearInventoryInternal(offlinePlayer, true, false);
+    }
+
+    public void clearInventoryInternal(OfflinePlayer offlinePlayer, boolean inventory, boolean enderchest) {
```
OmerBenGera: I'd expose the clearInventoryInternal method to NMSPlayers interface instead of seperating it into two different methods.
This is because when you want to clear both inventory and ender chest, you won't need to call this function twice (which loads data from disk twice for the player - expensive!)

src/main/java/com/bgsoftware/superiorskyblock/config/SettingsContainer.java
OmerBenGera, line 1: You don't convert old data to new format.
You should the values of `disband-inventory-clear` and `clear-on-join` and set the new fields accordingly.

Lemvs: I wanted to do this at first, but the two functions have completely different names - one of them has an inventory and the other doesn't, one has clear at the beginning, the other at the end, so without changing the names it would have messed up the names of these options 😔

OmerBenGera: > I wanted to do this at first, but the two functions have completely different names - one of them has an inventory and the other doesn't, one has clear at the beginning, the other at the end, so without changing the names it would have messed up the names of these options 😔

What do you mean?
Look for the `convertData` method in SettingsManagerImpl - you check if the old field exists, and if so change the data of the new fields accordingly

Lemvs:
@OmerBenGera is that what you meant? I'll try to solve the NMS problem in a moment, I wanted to fix this first

src/main/java/com/bgsoftware/superiorskyblock/nms/NMSPlayers.java
```diff
@@ -14,6 +14,8 @@ public interface NMSPlayers {
 
     void clearInventory(OfflinePlayer offlinePlayer);
```
OmerBenGera: You don't need this method anymore

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
```diff
@@ -1771,15 +1771,18 @@ public void setDescription(String description) {
...
         forEachIslandMember(Collections.emptyList(), false, islandMember -> {
             if (islandMember.equals(owner)) {
                 owner.setIsland(null);
             } else {
                 kickMember(islandMember);
             }
 
-            if (plugin.getSettings().isDisbandInventoryClear())
-                plugin.getNMSPlayers().clearInventory(islandMember.asOfflinePlayer());
+            if (clearInventory || clearEnderChest)
```
OmerBenGera: Unnecessary check

src/main/java/com/bgsoftware/superiorskyblock/commands/player/CmdAccept.java
```diff
@@ -101,8 +101,10 @@ public void execute(SuperiorSkyblockPlugin plugin, CommandSender sender, String[
 
         if (plugin.getSettings().isTeleportOnJoin())
             superiorPlayer.teleport(island);
-        if (plugin.getSettings().isClearOnJoin())
-            plugin.getNMSPlayers().clearInventory(superiorPlayer.asPlayer());
+        if (plugin.getSettings().isClearInventoryOnJoin() || plugin.getSettings().isClearEnderChestOnJoin())
```
OmerBenGera: Unnecessary check

src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminAdd.java
```diff
@@ -91,8 +91,10 @@ public void execute(SuperiorSkyblockPlugin plugin, CommandSender sender, @Nullab
 
         if (plugin.getSettings().isTeleportOnJoin() && targetPlayer.isOnline())
             targetPlayer.teleport(island);
-        if (plugin.getSettings().isClearOnJoin())
-            plugin.getNMSPlayers().clearInventory(targetPlayer.asOfflinePlayer());
+        if (plugin.getSettings().isClearInventoryOnJoin() || plugin.getSettings().isClearEnderChestOnJoin())
```
OmerBenGera: Unnecessary check

src/main/java/com/bgsoftware/superiorskyblock/nms/NMSPlayers.java
Lemvs: Ah, okay, I thought I shouldn't change the existing methods, so I did it this way from the beginning...

PR #2539, opened 2025-07-04
Extended "clear-on-..." options

Lemvs:


API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
```diff
@@ -282,9 +286,37 @@ public interface SettingsManager {
...
+     * Config-path: clear-inventory-on-disband
+     */
+    boolean isClearInventoryOnDisband();
+
+    /**
+     * Whether to clear players inventory when they join a new island or not.
+     * Config-path: clear-inventory-on-join
+     */
+    boolean isClearInventoryOnJoin();
+
```
OmerBenGera: Maybe instead of having boolean for each clear type, let's create an enum that represents different clear types (similar to how there are different respawn actions), and just have two methods:
```
List<ClearType> getOnJoinClearActions();
List<ClearType> getOnDisbandClearActions();
```
And don't make a real enum, make it similar to RespawnAction where you can register custom ClearTypes that will have custom actions

API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
```diff
@@ -285,37 +285,24 @@ public interface SettingsManager {
...
-    boolean isClearEnderChestOnDisband();
-
-    /**
-     * Whether to clear players Ender Chest when they join a new island or not.
-     * Config-path: clear-ender-chest-on-join
+     * Get the list of clear actions to perform on island members when their island is disbanded.
+     * Config-path: clear-on-disband
      */
-    boolean isClearEnderChestOnJoin();
+    List<ClearAction> getClearOnDisband();
```
OmerBenGera: `getClearActionsOnDisband`

API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
```diff
@@ -285,37 +285,24 @@ public interface SettingsManager {
...
-    boolean isClearInventoryOnDisband();
-
-    /**
-     * Whether to clear players inventory when they join a new island or not.
-     * Config-path: clear-inventory-on-join
+     * Get the list of clear actions to perform on players when they accept an invite to join an island.
+     * Config-path: clear-on-join
      */
-    boolean isClearInventoryOnJoin();
+    List<ClearAction> getClearOnJoin();
```
OmerBenGera: `getClearActionsOnJoin`

API/src/main/java/com/bgsoftware/superiorskyblock/api/player/inventory/ClearAction.java
OmerBenGera, line 1: A better design for this class is to have an abstract method that handles the clear for a player like so:
```public abstract void doClear(OfflinePlayer offlinePlayer);```

Then, instead of `clearInventory` of NMS to take a list of actions, you'll provide it with clearInventory and clearEnderChest booleans (or split the method, as you did before and I told you not to)

It might be better to think about a more optimized way to do it, as clearInventory is a bit heavy in performance

File: src/main/java/com/bgsoftware/superiorskyblock/config/SettingsContainer.java [hunk 4]
```diff
@@ -400,7 +400,24 @@ public SettingsContainer(SuperiorSkyblockPlugin plugin, YamlConfiguration config
...
+        this.clearOnDisbandActions = Collections.unmodifiableList(clearOnDisbandActions);
+        List<ClearAction> clearOnJoinActions = new LinkedList<>();
+        config.getStringList("clear-on-join").forEach(clearAction -> {
+            try {
+                clearOnJoinActions.add(ClearAction.getByName(clearAction));
+            } catch (NullPointerException error) {
+                Log.warnFromFile("config.yml", "Invalid clear on join action ", clearAction + ", skipping...");
+            }
+        });
+        this.clearOnJoinActions = Collections.unmodifiableList(clearOnJoinActions);
```
OmerBenGera: This can be moved into a separate function as it is duplicated code.
Especially if you want to add clear-on-leave option as well

API/src/main/java/com/bgsoftware/superiorskyblock/api/player/inventory/ClearAction.java
Lemvs: Hm, so how do I solve this, because for HEALTH, HUNGER etc. I also need to be able to load the offline player, so what's the best way to combine it?

OmerBenGera: > Hm, so how do I solve this, because for HEALTH, HUNGER etc. I also need to be able to load the offline player, so what's the best way to combine it?

Make each NMS have an interface they need to implement for accessing offline player data (Call it OfflinePlayerData or something)
Then, do something like this:
```
public void runClearActionsOnJoin(OfflinePlayer offlinePlayer) {
    OfflinePlayerData offlinePlayerData = plugin.getNMS().getOfflinePlayerData(offlinePlayer);
    OFFLINE_PLAYERS_DATA_CACHE.put(offlinePlayer.getUniqueId(), offlinePlayerData);
    for(ClearAction clearAction : ClearActions...) {
        clearAction.doClear(offlinePlayer);
    }
    OFFLINE_PLAYERS_DATA_CACHE.remove(offlinePlayer.getUniqueId());
    offlinePlayerData.saveChanges();
}
```
Can probably be more efficient to have `OFFLINE_PLAYERS_DATA_CACHE` be of type `ArrayMap`.
The `OfflinePlayerData` will have the methods you want to have - clearInventory, clearEnderChest, setHealth, setExp, etc.
Then in your doClear methods, retrieve the OfflinePlayerData and call the methods accordingly.
This way you don't need to expose each method in the NMS interface, and have everything handled by OfflinePlayerData

Lemvs: I think this is too advanced for me 😥

OmerBenGera: No worries, I'll do it and push to your branch.
Give me access to it

Lemvs: I invited you, thanks for help, I'm trying to learn something, but it's probably too much for me 🥲

src/main/java/com/bgsoftware/superiorskyblock/config/SettingsContainer.java
Lemvs: Fixed

API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
Lemvs: Fixed

API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
Lemvs: Fixed

API/src/main/java/com/bgsoftware/superiorskyblock/api/player/inventory/ClearAction.java
OmerBenGera: I pushed a commit with implementation for one nms version as I don't have more time today for the rest of the implementations.
From here it's a simple copy-paste across all the nms implementations and it should be good (needs testing obviously)

Lemvs: Okay, I'll look at it tomorrow, thank you 🥳

OmerBenGera:
> @OmerBenGera what do you think about moving the teleport-on-... option here? I would also add this option for create, kick, leave and call it actions-on-... If so, I would have to modify the island creation event a bit, I added teleport-on-create recently, then I also had to modify it - I added new constructor to api, so I can delete those constructor and orher changes, since it hasn't added in stable build yet? Because probably no one has yet used this new version of the event with their plugin 😅

I am against this change.

Lemvs:
I also added clear-on-kick, clear-on-leave and teleport-on-leave, because I had to separate the kickMember function anyway, and previously teleport-on-kick worked on both kick and leave. You can check if everything looks good

API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
```diff
@@ -251,8 +252,11 @@ public interface SettingsManager {
 
     /**
      * Whether inventory of island members should be cleared when their island is disbanded or not.
-     * Config-path: disband-inventory-clear
+     * Return true if clear-on-disband contains both ENDER_CHEST and INVENTORY.
+     * This method will be deleted in the future!
+     * @deprecated See {@link #getClearActionsOnDisband()}
      */
+    @Deprecated
     boolean isDisbandInventoryClear();
```
OmerBenGera, line 260: Should only check for `INVENTORY`

NMS/v1_12_R1/src/main/java/com/bgsoftware/superiorskyblock/nms/v1_12_R1/player/OfflinePlayerDataImpl.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public void applyChanges() {
+        this.fakePlayer.saveData();
+    }
+
+    @Override
+    public void release() {
+        this.fakePlayer = null;
```
OmerBenGera, line 66: You should call `POOL.release(this)` here
Do it for all versions

src/main/java/com/bgsoftware/superiorskyblock/config/SettingsManagerImpl.java
```diff
@@ -661,6 +693,24 @@ private void loadContainerFromConfig(YamlConfiguration cfg) throws ManagerLoadEx
     }
 
     private void convertData(YamlConfiguration cfg) {
+        if (cfg.contains("disband-inventory-clear")) {
+            if (cfg.getBoolean("disband-inventory-clear")) {
+                cfg.set("clear-on-disband", Arrays.asList("ENDER_CHEST", "INVENTORY"));
+            } else {
+                cfg.set("clear-on-disband", new ArrayList<>());
```
OmerBenGera: Use `Collections.emptyList()`

src/main/java/com/bgsoftware/superiorskyblock/config/SettingsManagerImpl.java
```diff
@@ -661,6 +693,24 @@ private void loadContainerFromConfig(YamlConfiguration cfg) throws ManagerLoadEx
     }
 
     private void convertData(YamlConfiguration cfg) {
+        if (cfg.contains("disband-inventory-clear")) {
```
OmerBenGera: Should be `cfg.isBoolean`

src/main/java/com/bgsoftware/superiorskyblock/core/menu/button/impl/LeaveButton.java
```diff
@@ -35,12 +35,8 @@ public void onButtonClick(InventoryClickEvent clickEvent) {
         SuperiorPlayer inventoryViewer = menuView.getInventoryViewer();
         Island island = inventoryViewer.getIsland();
 
-        if (getTemplate().leaveIsland && island != null && PluginEventsFactory.callIslandQuitEvent(island, inventoryViewer)) {
-            island.kickMember(inventoryViewer);
-
-            IslandUtils.sendMessage(island, Message.LEAVE_ANNOUNCEMENT, Collections.emptyList(), inventoryViewer.getName());
-
-            Message.LEFT_ISLAND.send(inventoryViewer);
+        if (getTemplate().leaveIsland) {
```
OmerBenGera:
```
if(getTemplate().leaveIsland && island != null) {
}
```

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
OmerBenGera, line 1: The best design would be to deprecate `kickMember` and add `removeMember` instead, which takes a new enum `KickReason` - that may be KICK and LEAVE for now.
Then, in the removeMember method, depends on the reason, you take the correct clear actions, teleport action, etc and perform it.

NMS/v1_12_R1/src/main/java/com/bgsoftware/superiorskyblock/nms/v1_12_R1/player/OfflinePlayerDataImpl.java
Lemvs: Should I do this instead of this.fakePlayer = null or in addition, if in addition then before or after?

OmerBenGera: > Should I do this instead of this.fakePlayer = null or in addition, if in addition then before or after?

In addition to setting the fakePlayer variable to null.
You can do it at the end of the function, not that it really matters.

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
Lemvs: Where should I create this enum? 😥

OmerBenGera: > Where should I create this enum? 😥

Under enums package in the API

Lemvs: > > Where should I create this enum? 😥
> 
> Under enums package in the API

That's exactly what I started doing, so I thought correctly 😀

API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
Lemvs: Hmm, but previously this option, like clear-on-join, also reset enderchest, so shouldn't it check both?

Lemvs:
I also added DISBAND to RemoveReason because kickMember was called in the disbandIsland function, which would cause both clear-on-disband and clear-on-kick actions to be performed

Lemvs:
I also added DISBAND to RemoveReason because kickMember was called in the disbandIsland function, which would cause both clear-on-disband and clear-on-kick actions to be performed, but these are different situations so only clear-on-disband actions should be performed then. And sorry for "Fix" and "Fix v2" but I was testing something and forgot to delete it 😅

API/src/main/java/com/bgsoftware/superiorskyblock/api/enums/RemoveReason.java
```diff
@@ -0,0 +1,9 @@
+package com.bgsoftware.superiorskyblock.api.enums;
+
+public enum RemoveReason {
```
OmerBenGera: `MemberRemoveReason` might be a better name? If you think about a better one, go ahead. `RemoveReason` is not indicative enough.
Also, please add comments for the enum class as well as its values.

API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
OmerBenGera: Then it should check both, right.

Lemvs:
> > I also added DISBAND to RemoveReason because kickMember was called in the disbandIsland function, which would cause both clear-on-disband and clear-on-kick actions to be performed, but these are different situations so only clear-on-disband actions should be performed then. And sorry for "Fix" and "Fix v2" but I was testing something and forgot to delete it 😅
> 
> 
> 
> Maybe we should add BANNED reason as well?

I also thought about it, but if you ban someone, you just remove them from the team and deny them entry to the island, and yet disband the island is a slightly different situation. In addition, you can also ban people who are not added, so adding the clear-on-ban option would be confusing 😕

OmerBenGera (approved):

Issue #2525, opened 2025-06-26
Need latest version of RoseStacker in maven repository to fix issue with stack deaths.


ValdemarF20:
### Minecraft's Version

Latest

### Plugin's Version

Latest

### Describe the bug

Stack entity death not working with superiorskyblock drop multiplier.

### To Reproduce

Kill stack of entities from RoseStacker with and without the upgrade of entity drops from SuperiorSkyblock, there is no difference when the entire stack dies at once. It works when one of the entities in the stack dies.

### Additional Information

The latest version of RoseStacker can be found here https://www.spigotmc.org/resources/rosestacker.82729/

I can fix the issue if that dependency is just uploaded to the bg-software maven repository.

OmerBenGera:
Added :)
Thanks for helping with bug fixing!

PR #2543, opened 2025-07-07
PR to fix https://github.com/BG-Software-LLC/SuperiorSkyblock2/issues/2525

ValdemarF20:


Hooks/RoseStacker/build.gradle
OmerBenGera, line 1: By doing this, it will no longer support older versions of RoseStacker.
What you should do is create a new module for newer versions of RoseStacker and only have the event registered in that module.

OmerBenGera (changes requested):

PR #2545, opened 2025-07-07
Added reset island settings option

Lemvs:


nivcoo:
I don't see it, does this feature is also available for /is admin command too ? If not i think it could be good to add it on it, thanks

OmerBenGera (approved):

Issue #2514, opened 2025-06-19
/is coop, and edit permissions via /is perms.


Aides123:
### Minecraft's Version

This server is running Leaf version 1.21.4-495-ver/1.21.4@91a911a (2025-06-18T17:14:16Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)
* You are running the latest version
Previous: 1.21.4-432-c1297d7 (MC: 1.21.4)

### Plugin's Version

#523 

### Describe the bug

Hey, as far as I'm concerned, it's a bug in the addition.

There is ‘/is coop’ but also /is setperms player has been added.
With the ‘/is setperms player’ command, a player can join multiple multiple islands as a ‘member’ which spoils the premise of /is coop.

My suggestion - fix this oversight.
Adding an option to disable /is setperms per player, and leave only roles.

### To Reproduce

I think the description above is accurate.

OmerBenGera:
You can disable the ability to change permissions by disabling the `/is permissions` command
Isn't this good enough?

Lemvs:
In my opinion, it would be useful to add separate permissions to:
/is balance [player]
/is counts [player]
/is permissions [player]
/is show [player]
/is team [player]
/is values ​​[player]

For example superior.island.xxx.player

I wanted to do this and even I implemented it, but the problem is that the getUsage method does not accept a player, which makes it impossible to check what permissions they have, so then this usage does not really fit. Is there any option to change this and add a method to the API that will also accept the player to be able to adjust the usage to his permissions, or maybe send what I did?

OmerBenGera:
> It would be enough if the menu with permissions continued to work. By disabling the command I disable permissions editing completely (Including from the GUI, unless I'm wrong?).
> 
> I want it to be possible to edit ONLY ROLES.

Gotcha - I added this option in latest dev build

> In my opinion, it would be useful to add separate permissions to: /is balance [player] /is counts [player] /is permissions [player] /is show [player] /is team [player] /is values ​​[player]
> 
> For example superior.island.xxx.player
> 
> I wanted to do this and even I implemented it, but the problem is that the getUsage method does not accept a player, which makes it impossible to check what permissions they have, so then this usage does not really fit. Is there any option to change this and add a method to the API that will also accept the player to be able to adjust the usage to his permissions, or maybe send what I did?
> 
[...]

PR #2549, opened 2025-07-09
Added separate permissions for commands that can be used on another island

Lemvs:
Fixes issue #2514 

Currently, I've made it so that if a player has the old permission, they can use both functions, just like before. If a player only has the .self permission, they can use the command without an argument, and if they only have the .other permission, they can't use the command at all (.other is just an extension of .self). I did it this way because you wanted to avoid having to change the granted permissions after introducing this function. However, I think this solution is quite confusing, and it would be better to separate these permissions – the old permission would grant access to the command without an argument, and the additional .other permission would allow the command to use an argument (this permission would also be an extension of the default permission, but it wouldn't be a children, but a separate permission). Let me know what you think.

### DIFF
--- a/API/src/main/java/com/bgsoftware/superiorskyblock/api/commands/SuperiorCommand.java
+++ b/API/src/main/java/com/bgsoftware/superiorskyblock/api/commands/SuperiorCommand.java
@@ -48,14 +48,6 @@ public interface SuperiorCommand {
     int getMinArgs();
 
     /**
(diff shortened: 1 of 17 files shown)


OmerBenGera:
Because there are many arguments for every method for the command now, I want to implement `CommandContext` class which will contain all the context for the command (executor, args, etc)
Therefore, I am closing this.

PR #2559, opened 2025-07-17
Added missing blocks to counts button

Lemvs:
Additionally, I sorted this map to make it more readable

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/core/menu/button/impl/CountsPagedObjectButton.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/core/menu/button/impl/CountsPagedObjectButton.java
@@ -181,7 +181,7 @@ public class CountsPagedObjectButton extends AbstractPagedMenuButton<MenuCounts.
             .put("RED_WALL_BANNER", "RED_BANNER")
             .put("REDSTONE_COMPARATOR_OFF", "REDSTONE_COMPARATOR")
             .put("REDSTONE_COMPARATOR_ON", "REDSTONE_COMPARATOR")
-            .put("REDSTONE_LAMP_ON", "REDSTONE_LAMP")
+            .put("REDSTONE_LAMP_ON", "REDSTONE_LAMP_OFF")
             .put("REDSTONE_TORCH_OFF", "REDSTONE_TORCH_ON")
             .put("REDSTONE_WALL_TORCH", "REDSTONE_TORCH")
             .put("REDSTONE_WIRE", "REDSTONE")


src/main/java/com/bgsoftware/superiorskyblock/core/menu/button/impl/CountsPagedObjectButton.java
```diff
@@ -67,151 +84,141 @@ public class CountsPagedObjectButton extends AbstractPagedMenuButton<MenuCounts.
...
+            .put("POWDER_SNOW_CAULDRON", "CAULDRON")
             .put("PUMPKIN_STEM", "PUMPKIN_SEEDS")
+            .put("PURPLE_CANDLE_CAKE", "PURPLE_CANDLE")
             .put("PURPLE_WALL_BANNER", "PURPLE_BANNER")
+            .put("RED_CANDLE_CAKE", "RED_CANDLE")
+            .put("RED_WALL_BANNER", "RED_BANNER")
             .put("REDSTONE_COMPARATOR_OFF", "REDSTONE_COMPARATOR")
             .put("REDSTONE_COMPARATOR_ON", "REDSTONE_COMPARATOR")
             .put("REDSTONE_LAMP_ON", "REDSTONE_LAMP")
-            .put("REDSTONE_TORCH_OFF", "REDSTONE_TORCH")
```
OmerBenGera, line 128: Does it work on old versions as well?

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/core/menu/button/impl/CountsPagedObjectButton.java
Lemvs: Yes, it was done wrong before, on legacy versions there was no such block

Lemvs: Yeah, this only applies to legacy versions and was previously done incorrectly - there was no such material as REDSTONE_TORCH. Now I noticed that the same problem also applies to REDSTONE_LAMP, so I fixed that as well.

Redstone materials for 1.8.8 and 1.12.2:
image

Lemvs: @OmerBenGera when will you have some time to check my PRs? I ask because I want to do more, but without adding some PR I can't continue some of them 😓

OmerBenGera:
It might be possible with the new API to just get the item's type out of a block's type.
Anyways, I am keeping this for now and merging this PR.

PR #2568, opened 2025-07-21
Fixed island names coloring

Lemvs:
- After disabling color support for island names, color codes were displayed unnecessarily in some places.
- After the administrator changed the island name, the message was not appropriate for the situation.
- After changing the minimum/maximum length of an island name in the configuration, this value also had to be changed in all language files, so I made this happen automatically.


OmerBenGera (changes requested):
I pushed some changes myself
Go through them and let me know if you see anything wrong that I did before I merge this PR

Lemvs:
> I pushed some changes myself Go through them and let me know if you see anything wrong that I did before I merge this PR

There are now problems with using island names in commands - the commands suggest island names without color codes, but if the island has a colored name, an error occurs when trying to use such a name

image
image

Lemvs:
> I pushed some changes myself Go through them and let me know if you see anything wrong that I did before I merge this PR

Additionally, if we change the name but the text is the same, only the colors will be different, we will receive a success message, but the name will not be changed - the colors remain the same. We should get a message that the name is the same and has not been changed

image

Lemvs:
> I pushed some changes myself Go through them and let me know if you see anything wrong that I did before I merge this PR

And setting a name with the same text but different colors via another island shouldn't be possible either, since the commands use names without colors

image

Lemvs:
> All the above should be fixed

Yes, it's okay now, you can merge

PR #2570, opened 2025-07-21
Fixed opening missions menu when there are per island missions

### LINKED ISSUE #2569: /is mission accessible without island
### Minecraft's Version

1.21.7

### Plugin's Version

SuperiorSkyblock2-2025.1-b548

### Describe the bug

Hello, when i've not island i'm able to do /island missions and it will display empty menu without any mission maybe you can add an message when player doesn't have island ? 

### To Reproduce

Disband your island and do /is mission

### Additional Information

I use also SSB Bridge idk if it's linked

Lemvs:
Fixes #2569

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/core/menu/button/impl/OpenMissionCategoryButton.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/core/menu/button/impl/OpenMissionCategoryButton.java
@@ -30,9 +30,7 @@ public class OpenMissionCategoryButton extends AbstractMenuViewButton<BaseMenuVi
 
     @Override
     public void onButtonClick(InventoryClickEvent clickEvent) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/bgsoftware/superiorskyblock/core/menu/button/impl/OpenMissionCategoryButton.java
```diff
@@ -28,6 +30,13 @@ public Template getTemplate() {
 
     @Override
     public void onButtonClick(InventoryClickEvent clickEvent) {
+        boolean requireIsland = getTemplate().missionCategory.getMissions().stream().allMatch(Mission::getIslandMission);
```
OmerBenGera: This value is computed every button click. Instead, it is better to cache it once per menu view.

OmerBenGera (approved):

PR #2579, opened 2025-07-29
Changed disband count option behavior to allow setting 0 disbands by default

Lemvs:
I've changed the disband-count option to default-disband-count, and its behavior. Previously, you couldn't set new players to have 0 disbands, because 0 meant the option was disabled. Now, the option can't be disabled, but -1 (or lower) means you have an infinite number of disbands. Disbands are only taken if the player has a positive amount of them. If the old option was set to 0 in the config, the new option will have a value of -1.

### DIFF
--- a/API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
+++ b/API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
@@ -190,7 +190,10 @@ public interface SettingsManager {
      * The default amount of disbands players receive when they first join the server.
      * If 0, then the disbands limit is disabled.
      * Config-path: disband-count
+     *
+     * @deprecated see {@link #getDefaultDisbandCount()}
      */
+    @Deprecated
     int getDisbandCount();
 
     /**
@@ -463,6 +466,13 @@ public interface SettingsManager {
      * Config-path: build-outside-island
      */
     boolean isBuildOutsideIsland();
+
+    /**
+     * The default number of disbands players receive when they first join the server.
+     * If set to -1, players will have unlimited disbands.
+     * Config-path: default-disband-count
+     */
+    int getDefaultDisbandCount();
 
     /**
      * The default language to be set for new players.
--- a/src/main/java/com/bgsoftware/superiorskyblock/commands/player/CmdDisband.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/commands/player/CmdDisband.java
@@ -67,7 +67,7 @@ public class CmdDisband implements IPermissibleCommand {
 
     @Override
     public void execute(SuperiorSkyblockPlugin plugin, SuperiorPlayer superiorPlayer, Island island, String[] args) {
-        if (!superiorPlayer.hasDisbands() && plugin.getSettings().getDisbandCount() > 0) {
+        if (!superiorPlayer.hasDisbands()) {
             Message.NO_MORE_DISBANDS.send(superiorPlayer);
             return;
         }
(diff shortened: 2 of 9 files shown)


wlm1455925277:
I like it, this is definitely a feature the plugin needs!

API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
```diff
@@ -464,6 +467,13 @@ public interface SettingsManager {
      */
     boolean isBuildOutsideIsland();
 
+    /**
+     * The default number of disbands players receive when they first join the server.
+     * If set to -1, players will have unlimited disbands.
+     * Config-path: default-disband-count
+     */
+    int getDefaultDisbandCount();
```
OmerBenGera: Even tho the config path was changed, there's no real need to change the code functions.

src/main/java/com/bgsoftware/superiorskyblock/commands/player/CmdDisband.java
```diff
@@ -67,7 +67,7 @@ public Message getPermissionLackMessage() {
 
     @Override
     public void execute(SuperiorSkyblockPlugin plugin, SuperiorPlayer superiorPlayer, Island island, String[] args) {
-        if (!superiorPlayer.hasDisbands() && plugin.getSettings().getDisbandCount() > 0) {
+        if (!superiorPlayer.hasDisbands()) {
```
OmerBenGera: What if a player doesn't have disbands, but then I disable them through the config file, so it is not required anymore?
You should keep the `getDisbandCount() > 0` check (change it to >=)

src/main/java/com/bgsoftware/superiorskyblock/commands/player/CmdDisband.java
```diff
@@ -84,7 +84,10 @@ public void execute(SuperiorSkyblockPlugin plugin, SuperiorPlayer superiorPlayer
                         island.getIslandBank().getBalance().multiply(BigDecimal.valueOf(BuiltinModules.BANK.disbandRefund))));
             }
 
-            superiorPlayer.setDisbands(superiorPlayer.getDisbands() - 1);
+            if (superiorPlayer.getDisbands() > 0) {
```
OmerBenGera: Should check if disbands are enabled, and only in this case subtract disband from the player (I know it wasn't there before)

src/main/java/com/bgsoftware/superiorskyblock/core/menu/button/impl/DisbandButton.java
```diff
@@ -48,7 +48,9 @@ public void onButtonClick(InventoryClickEvent clickEvent) {
                         targetIsland.getIslandBank().getBalance().multiply(BigDecimal.valueOf(BuiltinModules.BANK.disbandRefund))));
             }
 
-            inventoryViewer.setDisbands(inventoryViewer.getDisbands() - 1);
+            if (inventoryViewer.getDisbands() > 0) {
```
OmerBenGera: Similar check to config, and not the player's disbands

src/main/java/com/bgsoftware/superiorskyblock/player/SSuperiorPlayer.java
```diff
@@ -582,7 +580,7 @@ public void setDisbands(int disbands) {
 
     @Override
     public boolean hasDisbands() {
-        return disbands > 0;
+        return disbands != 0;
```
OmerBenGera: Should be reverted, as you no longer relay on that method

OmerBenGera (changes requested):

OmerBenGera:
Only one additional change needs to be made - remove `getDefaultDisbandCount` and switch back to using `getDisbandCount` (They are literally the same code-wise)

PR #2613, opened 2025-08-25
 Improved loading of data from yaml configuration files

Lemvs:
Plugin would not start when these options was set to "[ ]" due to using 'contains' instead of 'isConfigurationSection'


OmerBenGera (approved):

Lemvs:
> There are probably many more places where the plugin checks for `contains` instead of `isX` Do you want to look into all places and fix them as well?

Hmm, most of the "cfg.contains" are located in menu files and in SettingsManagerImpl and they are used to convert old files/settings to new ones. While I can test the conversion of some settings, I am not able to test the conversion of these menus, so should I change it there also?

OmerBenGera (approved):

PR #2624, opened 2025-09-02
Expanded island previews options

Lemvs:
# Changelog #
- Added saving of the game mode the player had before entering preview mode.
- Added the ability to change the default game mode in preview mode.
- Added maximum distance checking on the Y axis, previously only X and Z were checked.
- Added the ability to change the default distance a player can move before the preview mode is canceled.
- Added the ability to block commands in preview mode.

*Additionally, I fixed getVisitWarmup() because it was getting the value from home-warmup instead of visit-warmup 🤔 and other minor problems.*


src/main/java/com/bgsoftware/superiorskyblock/config/SettingsContainer.java
```diff
@@ -531,18 +535,21 @@ public SettingsContainer(SuperiorSkyblockPlugin plugin, YamlConfiguration config
         this.commandAliases = Collections.unmodifiableMap(commandAliases);
         valuableBlocks = KeySets.unmodifiableKeySet(
                 KeySets.createHashSet(KeyIndicator.MATERIAL, config.getStringList("valuable-blocks")));
-        Map<String, Location> islandPreviewLocations = new HashMap<>();
-        if (config.isConfigurationSection("preview-islands")) {
-            for (String schematic : config.getConfigurationSection("preview-islands").getKeys(false)) {
+        islandPreviewsGameMode = GameMode.valueOf(config.getString("island-previews.game-mode", "SPECTATOR"));
```
OmerBenGera: Should be String#toUpperCase(Locale.ENGLISH)

OmerBenGera: Should try-catch with IllegalArgumentException for invalid game modes

OmerBenGera (changes requested):

PR #2639, opened 2025-09-12
Disabled display of empty data in island info

Lemvs:
I disabled the display of empty data such as Discord, PayPal or members in the island information and corrected the information about visitors count - it should show the number of visitors, just like the name says, but showed unique visitors (I made it show both counts)

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/commands/player/CmdShow.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/commands/player/CmdShow.java
@@ -100,12 +100,14 @@ public class CmdShow implements ISuperiorCommand {
             infoMessage.append(Message.ISLAND_INFO_WORTH.getMessage(locale, island.getWorth())).append("\n");
         if (!Message.ISLAND_INFO_LEVEL.isEmpty(locale))
             infoMessage.append(Message.ISLAND_INFO_LEVEL.getMessage(locale, island.getIslandLevel())).append("\n");
-        if (!Message.ISLAND_INFO_DISCORD.isEmpty(locale) && island.hasPermission(sender, IslandPrivileges.DISCORD_SHOW))
+        if (!Message.ISLAND_INFO_DISCORD.isEmpty(locale) && !"None".equals(island.getDiscord()) && island.hasPermission(sender, IslandPrivileges.DISCORD_SHOW)) {
             infoMessage.append(Message.ISLAND_INFO_DISCORD.getMessage(locale, island.getDiscord())).append("\n");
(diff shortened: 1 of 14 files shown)


src/main/java/com/bgsoftware/superiorskyblock/commands/player/CmdShow.java
```diff
@@ -100,12 +100,14 @@ public void execute(SuperiorSkyblockPlugin plugin, CommandSender sender, String[
             infoMessage.append(Message.ISLAND_INFO_WORTH.getMessage(locale, island.getWorth())).append("\n");
         if (!Message.ISLAND_INFO_LEVEL.isEmpty(locale))
             infoMessage.append(Message.ISLAND_INFO_LEVEL.getMessage(locale, island.getIslandLevel())).append("\n");
-        if (!Message.ISLAND_INFO_DISCORD.isEmpty(locale) && island.hasPermission(sender, IslandPrivileges.DISCORD_SHOW))
+        if (!Message.ISLAND_INFO_DISCORD.isEmpty(locale) && !"None".equals(island.getDiscord()) && island.hasPermission(sender, IslandPrivileges.DISCORD_SHOW)) {
             infoMessage.append(Message.ISLAND_INFO_DISCORD.getMessage(locale, island.getDiscord())).append("\n");
-        if (!Message.ISLAND_INFO_PAYPAL.isEmpty(locale) && island.hasPermission(sender, IslandPrivileges.PAYPAL_SHOW))
+        }
+        if (!Message.ISLAND_INFO_PAYPAL.isEmpty(locale) && !"None".equals(island.getPaypal()) && island.hasPermission(sender, IslandPrivileges.PAYPAL_SHOW)) {
```
OmerBenGera, line 106: shouldn't it use the NONE message?

Lemvs: I thought about it, but if the island doesn't have Discord/PayPal set, then there's no need to show it - when the island doesn't have a name, it's not shown in the info either, so I decided to do the same here

OmerBenGera: > I thought about it, but if the island doesn't have Discord/PayPal set, then there's no need to show it - when the island doesn't have a name, it's not shown in the info either, so I decided to do the same here

In recent PRs you changed it so its no longer hardcoded to be "None", but a customizable message. If you check it against "None", it wont work if someone chnages the custom message

Lemvs: But in the previous PR I changed the display of "None" in the visitors/unique visitors menu, when the player does not have an island/the island does not have a name, I did not change anything with Discord/PayPal 😳

PR #2642, opened 2025-09-13
Added admin removeentitylimit command

Lemvs:
It was not possible to remove the entity limit for an island, so I added this option (based on remove block limit)


Lemvs:
> Add the new permission to the plugin.yml file


I did it in the second commit

OmerBenGera (approved):

PR #2645, opened 2025-09-13
Added the ability to use admin show command without an argument

Lemvs:
The command can now be used without an argument to check information about the island you are standing on, just like with the regular /is info

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminShow.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminShow.java
@@ -10,7 +10,7 @@ import com.bgsoftware.superiorskyblock.api.world.Dimension;
 import com.bgsoftware.superiorskyblock.api.world.WorldInfo;
 import com.bgsoftware.superiorskyblock.api.wrappers.SuperiorPlayer;
 import com.bgsoftware.superiorskyblock.commands.CommandTabCompletes;
-import com.bgsoftware.superiorskyblock.commands.ISuperiorCommand;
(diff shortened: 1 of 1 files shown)


src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminShow.java
```diff
@@ -38,7 +40,7 @@
 import java.util.function.Function;
 import java.util.function.Supplier;
 
-public class CmdAdminShow implements IAdminIslandCommand {
+public class CmdAdminShow implements ISuperiorCommand {
```
OmerBenGera: Should implement `IAdminIslandCommand`

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminShow.java
Lemvs: Then this change makes no sense because it will force you to provide an argument

OmerBenGera: > Then this change makes no sense because it will force you to provide an argument

You can override the general execute method, and not the one that takes a target argument

OmerBenGera: If you don't make it `IAdminIslandCommand`, people will be able to disable this command in the config file, and you should not be able to do so

OmerBenGera (approved):

PR #2648, opened 2025-09-14
Added raid trigger privilege

Lemvs:
Instead of allowing only island members to trigger raids, I added a privilege that allows you to choose which role can trigger raids, as suggested in this [Discussion](https://github.com/BG-Software-LLC/SuperiorSkyblock2/discussions/2610) (so after merging, close it)

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/island/privilege/IslandPrivileges.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/island/privilege/IslandPrivileges.java
@@ -68,6 +68,8 @@ public class IslandPrivileges {
     @Nullable
     public static final IslandPrivilege PICKUP_LECTERN_BOOK = register("PICKUP_LECTERN_BOOK", ServerVersion.isAtLeast(ServerVersion.v1_14));
     public static final IslandPrivilege PROMOTE_MEMBERS = register("PROMOTE_MEMBERS", IslandPrivilege.Type.COMMAND);
+    @Nullable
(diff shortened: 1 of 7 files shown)


Lemvs:
> It should probably be the same privilege as attacking mobs. If you can attack mobs, you should be able to trigger raids

Hmm, but this causes spawn of mobs + ravenger can destroy blocks, so I think that combining it with the "Monster Damage" privilege is too much of a simplification, players may want to invite others to mob grinders, but not allow them to trigger raids.

OmerBenGera:
> > It should probably be the same privilege as attacking mobs. If you can attack mobs, you should be able to trigger raids
> 
> Hmm, but this causes spawn of mobs + ravenger can destroy blocks, so I think that combining it with the "Monster Damage" privilege is too much of a simplification, players may want to invite others to mob grinders, but not allow them to trigger raids.

Currently, visitors cannot trigger raids.
Are you sure this is necessary and people will actually use it?

Lemvs:
> > > It should probably be the same privilege as attacking mobs. If you can attack mobs, you should be able to trigger raids
> > 
> > 
> > Hmm, but this causes spawn of mobs + ravenger can destroy blocks, so I think that combining it with the "Monster Damage" privilege is too much of a simplification, players may want to invite others to mob grinders, but not allow them to trigger raids.
> 
> Currently, visitors cannot trigger raids. Are you sure this is necessary and people will actually use it?


Someone started this discussion, so I decided to introduce it 😕

PR #2649, opened 2025-09-14
Added tamed animal damage privilege

Lemvs:
If players want to invite others to animal spawners, they can't have tamed animals on their island because the guests will be able to kill them, so I added a separate permission for tamed animal damage to prevent this situation.


src/main/java/com/bgsoftware/superiorskyblock/service/region/RegionManagerServiceImpl.java
```diff
@@ -320,7 +320,8 @@ public InteractionResult handleEntityDamage(Entity damager, Entity entity) {
         try (ObjectsPools.Wrapper<Location> wrapper = ObjectsPools.LOCATION.obtain()) {
             Location entityLocation = entity.getLocation(wrapper.getHandle());
             Island island = plugin.getGrid().getIslandAt(entityLocation);
-            IslandPrivilege islandPrivilege = BukkitEntities.getCategory(entity.getType()).getDamagePrivilege();
+            IslandPrivilege islandPrivilege = BukkitEntities.isTameable(entity) ? IslandPrivileges.TAMED_ANIMAL_DAMAGE :
```
OmerBenGera, line 349: This check should be in the EntityCategory#getDamagePrivilege function

OmerBenGera (changes requested):

OmerBenGera:
Also, fix the conflicts

src/main/java/com/bgsoftware/superiorskyblock/service/region/RegionManagerServiceImpl.java
Lemvs: > This check should be in the EntityCategory#getDamagePrivilege function


In that file, EntityTypes are used, and checking if a mob is tamed requires a specific Entity, not its type

OmerBenGera (approved):

PR #2693, opened 2025-10-05
improv. fix spawn without breeder

nivcoo:
Hello

I improved the way to retrieve entity spawn.

I guess that we have to add more event on it, like dispense event or creature spawn like iron golem or Bucket fish etc but i'm not sure.

I was not able to test it yet, because i don't have a test server with island world (i use SSB bridge so i've to create another server without it).

If you have any suggestion let me know


nivcoo:
I had to leave until Wednesday I don’t check if I forget something to edit yet and didn’t tried it too I will do it when I will come back

PR #2707, opened 2025-10-10
Added option to create a separate item for locked missions

Lemvs:
# Changelog # 
- I added option to create a locked icon and sound, so you can set a separate item for missions for which you haven't completed the required missions or don't pass the required checks. This feature is optional, so if someone doesn't add a locked sound or icon, the items will be created as before
- I fixed ability to click in the mission menu when you don't have an island, as missions can be completed without an island, but clicking in the menu was blocked
- I simplified replacing placeholders in mission icons to avoid duplicating the same code multiple times

### DIFF
--- a/src/main/resources/modules/missions/categories/fisherman/fisherman_31_13.yml
+++ b/src/main/resources/modules/missions/categories/fisherman/fisherman_31_13.yml
@@ -52,7 +52,6 @@ icons:
       - '&8 - &7$50,000'
       - ''
       - '&6Treasure Items: &7{1}/3'
-      - '&6Progress: &7{0}%'
       - '&c&l ✘ &7Locked'
     amount: 3
   not-completed:
@@ -69,6 +68,7 @@ icons:
       - '&8 - &7$50,000'
       - ''
       - '&6Treasure Items: &7{1}/3'
+      - '&6Progress: &7{0}%'
       - '&c&l ✘ &7Not Completed'
     amount: 3
   can-complete:


src/main/java/com/bgsoftware/superiorskyblock/core/itemstack/ItemBuilder.java
```diff
@@ -191,6 +192,13 @@ public ItemBuilder replaceAll(String regex, String replace) {
         return this;
     }
 
+    public ItemBuilder replaceAll(Map<String, String> map) {
+        for (Map.Entry<String, String> entry : map.entrySet()) {
+            replaceAll(entry.getKey(), entry.getValue());
+        }
+        return this;
+    }
+
```
OmerBenGera: I don't like this

src/main/java/com/bgsoftware/superiorskyblock/core/menu/button/impl/MissionsPagedObjectButton.java
```diff
@@ -79,23 +98,25 @@ public ItemStack modifyViewItem(ItemStack buttonItem) {
...
-                                .build(inventoryViewer) :
-                        missionData.getNotCompleted()
-                                .replaceAll("{0}", percentage + "")
-                                .replaceAll("{1}", progressValue + "")
-                                .replaceAll("{2}", amountCompleted + "")
-                                .build(inventoryViewer);
+        Map<String, String> placeholders = new HashMap<>();
+        placeholders.put("{0}", percentage + "");
+        placeholders.put("{1}", progressValue + "");
+        placeholders.put("{2}", amountCompleted + "");
```
OmerBenGera: Instead of this, just keep a reference to ItemBuilder, then replace the placeholders and build the itemstack only once

src/main/resources/modules/missions/categories/farmer/farmer_2.yml
```diff
@@ -44,6 +44,26 @@ required-plants:
...
+      - '&8 - &7x10 Melons'
+      - ''
+      - '&6Rewards:'
+      - '&8 - &7x1 Sugar Cane'
+      - '&8 - &7x1 Cactus'
+      - '&8 - &7$5,000'
+      - ''
+      - '&6Grown Pumpkins: &7{value_pumpkin}/10'
+      - '&6Grown Melons: &7{value_melon}/10'
+      - '&6Progress: &7{0}%'
```
OmerBenGera: I want to say that it shouldn't show progress
Other than that, looks good

Lemvs: So just delete this line for locked icons? Or set it to 0%?

OmerBenGera: > So just delete this line for locked icons? Or set it to 0%?

Delete, similar to how "completed" icons don't have it

Lemvs: But completed icons have progress, it just says 100%

image

OmerBenGera: > But completed icons have progress, it just says 100%
> 
> image

I think it should be hidden, as it is "locked"

PR #2708, opened 2025-10-10
Fixed /is admin setpermission command

Lemvs:
# Changelog # 
- Tab completes did not show roles for COMMAND permissions, and the command completely prevented setting roles for this permission type.
- Tab completes did not show the guest and coop roles for ACTION permissions, and the command completely prevented setting these roles for this permission type.

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminSetPermission.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminSetPermission.java
@@ -98,20 +98,22 @@ public class CmdAdminSetPermission implements IAdminIslandCommand {
 
     @Override
     public List<String> adminTabComplete(SuperiorSkyblockPlugin plugin, CommandSender sender, Island island, String[] args) {
-        if (args.length == 4) {
-            return CommandTabCompletes.getIslandPrivileges(args[3]);
-        } else if (args.length == 5) {
-            IslandPrivilege islandPrivilege = getIslandPrivilegeSafe(args[3]);
+        switch (args.length) {
+            case 4:
+                return CommandTabCompletes.getIslandPrivileges(args[3]);
+            case 5: {
+                IslandPrivilege islandPrivilege = getIslandPrivilegeSafe(args[3]);
 
-            if (islandPrivilege != null) {
-                if (islandPrivilege.getType() == IslandPrivilege.Type.COMMAND)
+                if (islandPrivilege == null)
+                    return Collections.emptyList();
+                else if (islandPrivilege.getType() == IslandPrivilege.Type.COMMAND)
                     return CommandTabCompletes.getPlayerRoles(plugin, args[4], PlayerRole::isRoleLadder);
                 else
-                    return CommandTabCompletes.getPlayerRoles(plugin, args[4], playerRole -> true);
+                    return CommandTabCompletes.getPlayerRoles(plugin, args[4], null);
             }
+            default:
+                return Collections.emptyList();
         }
-
-        return Collections.emptyList();
     }
 
     @Nullable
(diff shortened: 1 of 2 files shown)


src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminSetPermission.java
```diff
@@ -97,17 +98,20 @@ else if (targetPlayer == null)
...
     @Override
     public List<String> adminTabComplete(SuperiorSkyblockPlugin plugin, CommandSender sender, Island island, String[] args) {
-        switch (args.length) {
-            case 4:
-                return CommandTabCompletes.getIslandPrivileges(args[3]);
-            case 5: {
-                IslandPrivilege islandPrivilege = getIslandPrivilegeSafe(args[3]);
-                return islandPrivilege == null || islandPrivilege.getType() == IslandPrivilege.Type.COMMAND ?
-                        Collections.emptyList() : CommandTabCompletes.getPlayerRoles(plugin, args[4], PlayerRole::isRoleLadder);
+        if (args.length == 4) {
```
OmerBenGera: What's wrong with switch-case?

src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminSetPermission.java
```diff
@@ -97,17 +98,20 @@ else if (targetPlayer == null)
...
+            return CommandTabCompletes.getIslandPrivileges(args[3]);
+        } else if (args.length == 5) {
+            IslandPrivilege islandPrivilege = getIslandPrivilegeSafe(args[3]);
+
+            if (islandPrivilege != null) {
+                if (islandPrivilege.getType() == IslandPrivilege.Type.COMMAND)
+                    return CommandTabCompletes.getPlayerRoles(plugin, args[4], PlayerRole::isRoleLadder);
+                else
+                    return CommandTabCompletes.getPlayerRoles(plugin, args[4], playerRole -> true);
```
OmerBenGera: Make the predicate null

src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminSetPermission.java
Lemvs: Because I didn't add default and there are only 2 cases, and when I did it recently, codacy recognized it as an issue :(

src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminSetPermission.java
Lemvs: getPlayerRoles, unlike getOnlinePlayersAndIslands, does not skip the predicate when it is null, so will this work?

src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminSetPermission.java
OmerBenGera: Then add the default switch-case. You return Collections.emptyList().

src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminSetPermission.java
OmerBenGera: > getPlayerRoles, unlike getOnlinePlayersAndIslands, does not skip the predicate when it is null, so will this work?

No, it should be changed to allow null predicates. This is more optimized than running predicates that always return true

src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminSetPermission.java
Lemvs: But then I need to return the same thing when islandPrivilege is null, so I converted it to return it only once

PR #2731, opened 2025-10-22
Added message-delays option

Lemvs:
I decided to implement the option mentioned in this [Discussion](https://github.com/BG-Software-LLC/SuperiorSkyblock2/discussions/2730), but instead of creating another collection with delays only for `MISSION_NO_AUTO_REWARD` message, as is currently the case for `SCHEMATIC` and `PROTECTION` messages, I decided to allow players to create their own list of messages that should have delays.

By introducing this option, I can transfer the functionality of `protected-message-delay` option and add ability to set own delays for `SCHEMATIC`, `BUILD_OUTSIDE_ISLAND`, and `ISLAND_BEING_CALCULATED`. 

The only problem is that I can't use the methods I added in `PROTECTION` and `SCHEMATIC` messages, so maybe instead of creating such a custom message, I'll just send these messages normally? I'll simply move the `PROTECTION` functionality to `ProtectionHelper` and use the messages directly in `PortalsManagerServiceImpl` instead of through SCHEMATIC. And in the config, I would set by default the delays for these messages as they were before.

What do you think about this @OmerBenGera ?

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/config/SettingsManagerImpl.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/config/SettingsManagerImpl.java
@@ -711,7 +711,7 @@ public class SettingsManagerImpl extends Manager implements SettingsManager {
     }
 
     private void convertData(YamlConfiguration cfg) {
-        if (cfg.isLong("protected-message-delay")) {
+        if (cfg.get("protected-message-delay") instanceof Number) {
             long delay = cfg.getLong("protected-message-delay") * 50;
             cfg.set("message-delays.ISLAND_PROTECTED", delay);
             cfg.set("message-delays.ISLAND_PROTECTED_OPPED", delay);


src/main/java/com/bgsoftware/superiorskyblock/config/SettingsContainer.java
```diff
@@ -562,6 +563,13 @@ public SettingsContainer(SuperiorSkyblockPlugin plugin, YamlConfiguration config
         tabCompleteHideVanished = config.getBoolean("tab-complete-hide-vanished", true);
         dropsUpgradePlayersMultiply = config.getBoolean("drops-upgrade-players-multiply", false);
         protectedMessageDelay = config.getLong("protected-message-delay", 60L);
+        Map<String, Long> messageDelays = new HashMap<>();
+        if (config.isConfigurationSection("message-delays")) {
+            for (String message : config.getConfigurationSection("message-delays").getKeys(false)) {
+                messageDelays.put(message, config.getLong("message-delays." + message));
+            }
+        }
+        this.messageDelays = messageDelays;
```
OmerBenGera: Collections.unmodifiableMap

src/main/java/com/bgsoftware/superiorskyblock/core/messages/Message.java
```diff
@@ -1079,6 +1085,27 @@ private void setMessage(Locale locale, IMessageComponent messageComponent) {
         messages.put(locale, messageComponent);
     }
 
+    private long getDelay() {
+        Long delay = plugin.getSettings().getMessageDelays().get(this.name());
+        return delay == null ? 0L : delay;
+    }
+
+    private boolean addDelay(UUID playerUUID) {
+        if (getDelay() > 0L) {
```
OmerBenGera: Cache the result of `getDelay`

src/main/java/com/bgsoftware/superiorskyblock/core/messages/Message.java
```diff
@@ -1079,6 +1085,27 @@ private void setMessage(Locale locale, IMessageComponent messageComponent) {
...
+            if (delays == null) {
+                delays = new HashMap<>();
+            }
```
OmerBenGera: delays should not be null. Also, delays should be ConcurrentHashMap

src/main/java/com/bgsoftware/superiorskyblock/core/messages/Message.java
```diff
@@ -1095,6 +1122,7 @@ public static void registerListeners(PluginEventsDispatcher dispatcher) {
 
     private static void onSettingsUpdate() {
         noInteractMessages = null;
+        delays = null;
```
OmerBenGera: Why?

src/main/java/com/bgsoftware/superiorskyblock/core/messages/Message.java
```diff
@@ -907,6 +908,8 @@ protected MessagesServiceImpl create() {
 
     @Nullable
     private static Collection<UUID> noInteractMessages;
```
OmerBenGera: Should not be used anymore, as well as noSchematicMessages

src/main/java/com/bgsoftware/superiorskyblock/config/SettingsManagerImpl.java
```diff
@@ -585,6 +585,11 @@ public long getProtectedMessageDelay() {
         return this.global.getProtectedMessageDelay();
```
OmerBenGera: Should be deprecated

src/main/java/com/bgsoftware/superiorskyblock/config/SettingsContainer.java
```diff
@@ -562,6 +563,13 @@ public SettingsContainer(SuperiorSkyblockPlugin plugin, YamlConfiguration config
         tabCompleteHideVanished = config.getBoolean("tab-complete-hide-vanished", true);
         dropsUpgradePlayersMultiply = config.getBoolean("drops-upgrade-players-multiply", false);
         protectedMessageDelay = config.getLong("protected-message-delay", 60L);
```
OmerBenGera: No need for that anymore

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/core/messages/Message.java
Lemvs: Some time ago you did this for noInteractMessages, because after changing the delay in the config and reloading the plugin it wasn't updating, so I thought it was necessary to do the same here 🤔

Lemvs:
I also added a SCHEMATIC_NOT_ADDED message so that you can set a delay for it and translate it as well.
In the default config I entered all the messages that previously had delays.

But there is one problem, in SchematicManagerImpl in the convertData method I added converting the old protected-message-delay value, but it doesn't work - the values ​​are not changed, do you know why @OmerBenGera ?

I added it to IGNORED_SECTIONS, because without it, when loading the config, the deleted default messages from message-delays were coming back, I don't know if that has anything to do with it

src/main/java/com/bgsoftware/superiorskyblock/core/messages/Message.java
```diff
@@ -906,7 +849,7 @@ protected MessagesServiceImpl create() {
     };
 
     @Nullable
-    private static Collection<UUID> noInteractMessages;
+    private static Map<Message, Collection<UUID>> delays;
```
OmerBenGera: You don't need this, you already have Message#delayedMessages (which can be renamed to just delays)

src/main/java/com/bgsoftware/superiorskyblock/core/messages/Message.java
```diff
@@ -1064,6 +1003,9 @@ public void send(CommandSender sender, Locale locale, Object... args) {
             UUID playerUUID = ((Player) sender).getUniqueId();
             if (delayedMessages != null && !delayedMessages.add(playerUUID))
                 return;
+            if (addDelay(playerUUID)) {
+                return;
+            }
```
OmerBenGera: Redundant. You can use `delayedMessages`

src/main/java/com/bgsoftware/superiorskyblock/core/messages/Message.java
```diff
@@ -1079,6 +1021,28 @@ private void setMessage(Locale locale, IMessageComponent messageComponent) {
...
+            Collection<UUID> collection = delays.computeIfAbsent(this, message ->
+                    AutoRemovalCollection.newHashSet(delay, TimeUnit.MILLISECONDS)
+            );
+
+            return !collection.add(playerUUID);
+        }
+
+        return false;
+    }
+
```
OmerBenGera: Can be removed probably

src/main/java/com/bgsoftware/superiorskyblock/core/messages/Message.java
```diff
@@ -1094,7 +1058,7 @@ public static void registerListeners(PluginEventsDispatcher dispatcher) {
     }
 
     private static void onSettingsUpdate() {
-        noInteractMessages = null;
+        delays = null;
```
OmerBenGera: Instead, you should go through the message delayes and initialize delayedMessages for each Message:
for(Message message : Message.values()) {
    message.delatedMessages = plugin.getSettings().getDelayedMessages()contains(message.name) ? Auto removal map : null
}

src/main/java/com/bgsoftware/superiorskyblock/service/portals/PortalsManagerServiceImpl.java
```diff
@@ -130,14 +129,14 @@ private EntityPortalResult handlePlayerPortalInternal(SuperiorPlayer superiorPla
...
+        if (portalResult == EntityPortalResult.WORLD_NOT_UNLOCKED) {
             Dimension originalDestination = getTargetWorld(portalLocation, portalType);
-            Message.SCHEMATICS.send(superiorPlayer, Message.WORLD_NOT_UNLOCKED.getMessage(
-                    superiorPlayer.getUserLocale(), Formatters.CAPITALIZED_FORMATTER.format(originalDestination.getName())));
+            Message.WORLD_NOT_UNLOCKED.send(superiorPlayer, Formatters.CAPITALIZED_FORMATTER.format(originalDestination.getName()));
         }
 
-        if (portalResult == EntityPortalResult.DESTINATION_WORLD_DISABLED && !Message.WORLD_NOT_ENABLED.isEmpty(superiorPlayer.getUserLocale())) {
-            Message.SCHEMATICS.send(superiorPlayer, Message.WORLD_NOT_ENABLED.getMessage(superiorPlayer.getUserLocale()));
+        if (portalResult == EntityPortalResult.DESTINATION_WORLD_DISABLED) {
```
OmerBenGera: else if / switch

src/main/java/com/bgsoftware/superiorskyblock/service/region/ProtectionHelper.java
```diff
@@ -33,4 +46,26 @@ public static boolean shouldPreventInteraction(InteractionResult interactionResu
         throw new IllegalStateException("No handling for result " + interactionResult);
     }
 
+    public static void sendProtectionMessage(CommandSender sender) {
```
OmerBenGera, line 49: In most cases (if not all of them), Island was already looked-up by the caller.
Add a nullable island parameter to the method. Only if it's null, look for the island again.

OmerBenGera (changes requested):

OmerBenGera:
> I also added a SCHEMATIC_NOT_ADDED message so that you can set a delay for it and translate it as well. In the default config I entered all the messages that previously had delays.
> 
> But there is one problem, in SchematicManagerImpl in the convertData method I added converting the old protected-message-delay value, but it doesn't work - the values ​​are not changed, do you know why @OmerBenGera ?
> 
> I added it to IGNORED_SECTIONS, because without it, when loading the config, the deleted default messages from message-delays were coming back, I don't know if that has anything to do with it

Regarding convertData - I assume you meant SettingsManagerImpl. It should work, can't think about a reason why it wouldn't.
You need to add that section to the ignored sections, otherwise it will be synced with defaults. That's correct.

src/main/java/com/bgsoftware/superiorskyblock/service/region/ProtectionHelper.java
Lemvs: Hmm, but shouldPreventInteraction(), which uses this method, doesn't take an island as an argument, so I should add it there and correct its use in other files?

OmerBenGera: > In most cases (if not all of them), Island was already looked-up by the caller. Add a nullable island parameter to the method. Only if it's null, look for the island again.

Yep

Lemvs: But the shouldPreventInteraction() method is mainly used in ProtectionListener, where the island is not known yet, only in RegionManagerServiceImpl there are islands, so how to connect it? Should I "assign" the island to InteractionResult or how? 😕

OmerBenGera: > But the shouldPreventInteraction() method is mainly used in ProtectionListener, where the island is not known yet, only in RegionManagerServiceImpl there are islands, so how to connect it? Should I "assign" the island to InteractionResult or how? 😕

In this case, pass null

Lemvs: > > But the shouldPreventInteraction() method is mainly used in ProtectionListener, where the island is not known yet, only in RegionManagerServiceImpl there are islands, so how to connect it? Should I "assign" the island to InteractionResult or how? 😕
> 
> In this case, pass null

But any method that uses `ProtectionHelper.shouldPreventInteraction()` in `ProtectionListener` doesn't know the island, so what's the point of providing it since I will have to use null everywhere? Only `RegionManagerServiceImpl` know island, but it only returns `InteractionResult` to `ProtectionListener`. 

To be able to give an island to `ProtectionHelper`, you would have to use it directly in `RegionManagerServiceImpl`, instead of `ProtectionListener`, or add the island field to `InteractionResult`

OmerBenGera: Gotcha, so let's not add the parameter for now

src/main/java/com/bgsoftware/superiorskyblock/core/messages/Message.java
```diff
@@ -1094,7 +1029,19 @@ public static void registerListeners(PluginEventsDispatcher dispatcher) {
...
+
+            if (delay > 0L)
+                message.delayedMessages = AutoRemovalCollection.newHashSet(delay, TimeUnit.MILLISECONDS);
+            else
+                message.delayedMessages = null;
+        }
+    }
+
+    private long getDelay() {
+        Long delay = plugin.getSettings().getMessageDelays().get(this.name());
```
OmerBenGera: getOrDefault

OmerBenGera (changes requested):

PR #2742, opened 2025-10-30
feat. add unified schema to save/load mission save

nivcoo:
Hello,

I would add new datas on saved file like : start-member-count but all missions had his own scehma, i unified all, actually i did changes only for BlockMissions to check with you if SSB code is OK and after i will do same for all type.

I deprecated old one, and make sure that it still works.


OmerBenGera:
I don't understand.
You can save into the section whatever you want in the mission.
What's the purpose of making all missions save data in the same scheme?

nivcoo:
I want to implement feature like : Add amount per member for all mission, and with another option to based the amount on the number of member when the mission start, but actually we can't, maybe we can for BlockMissions because you use "counts" name but all other use different schema, so it will not possible to do that

nivcoo:
With my changes, all missions saves will be migrated automaticaly with the new format, that mean in 1 year or 2, you will remove old migrated methods and deprecated one

nivcoo:
So, 

I want to add the feature : per-member-add-amount, like that we will be able to add more difficulty if there is more member in island, but players aren't stupid, if we don't add protection they will just farm with 8 players until the amount of 1 player in island and just leave to have rewards and join after, to protect that i want to add an option in same quest to choose if we what to use the start-members-count, like that even if player leave the amount required will be the same as before, but to do that i need to save this variable somewhere, and actually i can't due to the way the datas are saved

nivcoo:
And it's not to create external mission, it's to edit current mission type to support the "per_membercount" feature

nivcoo:
My goal is :

Edit all default mission type, to support the per-member-amount feature, but to do that i have to unified the way the quest progress is actually saved, that's why I PR this, to unified the datas and migrate old format, and after add the new feature in another PR

nivcoo:
You are right i can juste create custom jar but if I do that the feature will not be available in default missions types

OmerBenGera:
> You are right i can juste create custom jar but if I do that the feature will not be available in default missions types

This PR doesnt add this feature neither...

nivcoo:
If I don’t have it say me how can I say member-start-count for fishing mission so ?

OmerBenGera:
> If I don’t have it say me how can I say member-start-count for fishing mission so ?

In the saveProgress method:
section.set("member-start-count", <value>);

Then in loadProgress method:
section.get("member-start-count");

Its a yaml file, save there whatever you want

nivcoo:
No i can't do that look my screen, 
image
how can i add this variable here ?

nivcoo:
That's i the problem, I can save the variable only for blocks mission because the counts is under "counts" section, others mission the list is under island UUID

OmerBenGera:
> That's i the problem, I can save the variable only for blocks mission because the counts is under "counts" section, others mission the list is under island UUID

Just make a section for your own data...

OmerBenGera:
> Yes so save 2 times the uuid thats why its not the best things

Then use:
section.set(uuid + ".member-start-count", <value>);

nivcoo:
> > Yes so save 2 times the uuid thats why its not the best things
> 
> Then use: section.set(uuid + ".member-start-count", );

But it will break the list '0' '1' etc no ?

PR #2754, opened 2025-11-01
Renamed level formula option

Lemvs:
1. I changed the name of the `island-level-formula` option to `block-level-formula` because the old name was misleading; the formula is used to calculate the block level, not the island level. I only kept the old name in the API, as it's not a significant change.
2. I removed the script engine installation information from the description of this option, because according to that link, this process happens automatically anyway.
3. In SettingsManagerImpl, I didn't use isString() in convertData() because {} prevented it from being considered a String. The ability to use only numbers also created this problem, and you didn't want to use contains(), so I used get() and compared it to null.

### DIFF
--- a/API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
+++ b/API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
@@ -78,9 +78,9 @@ public interface SettingsManager {
     StackedBlocks getStackedBlocks();
 
     /**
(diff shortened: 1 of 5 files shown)


Lemvs:
> Why not using "isString"? It is always a string.

Read 3. It's not always a String, I checked it and isString() didn't work. From what I've found, {} is considered an object/map (?), and that causes problems. Using just a number also causes problems, because it's written in '', not ""

OmerBenGera (approved):

PR #2762, opened 2025-11-05
Added the ability to open mission categories via command

Lemvs:
I added the ability to specify a mission category in the /is missions command so that you can open it directly without opening it through the main menu.
(Someone probably even planned to introduce such an option, because minArgs was set to 1, maxArgs to 2 while this command had no arguments 😳)

Additionally, I fixed the mission suggestions in commands, because if the names are in upper case, tab completes did not work correctly.


src/main/java/com/bgsoftware/superiorskyblock/module/missions/commands/CmdMissions.java
```diff
@@ -52,18 +55,42 @@ public boolean canBeExecutedByConsole() {
...
         SuperiorPlayer superiorPlayer = plugin.getPlayers().getSuperiorPlayer(sender);
-        boolean requireIsland = plugin.getMissions().getAllMissions().stream().allMatch(Mission::getIslandMission);
 
-        if (requireIsland && !superiorPlayer.hasIsland()) {
-            Message.INVALID_ISLAND.send(superiorPlayer);
-            return;
-        }
+        if (args.length == 1) {
+            boolean requireIsland = plugin.getMissions().getAllMissions().stream().allMatch(Mission::getIslandMission);
+
```
OmerBenGera: This value can be cached in MissionsManager and be updated when the plugin reloads (SETTINGS_UPDATE_EVENT will work for that probably)

OmerBenGera (changes requested):

Lemvs:
I did something like this, but there's a problem: opening a category via menu without having an island only works after a plugin reload, while by /is missions <category> work without reload. Additionaly, changing mission types between player/island requires a double plugin reload. When I previously storing categories in the set, rather than their names, caused even more problems, opening via menu completny doesn't work - this always returned that the category was not in the set, which made requireIsland true for every category, no matter what missions it contained.

How can I solve this? Missions are stored in the MissionsContainer, but it's not static in MissionsManagerImpl, so I wanted to access it from a plugin, but perhaps that's causing the problem, as it might not be updated at this stage? @OmerBenGera

OmerBenGera:
> ing a category via menu without having an island only works after a plugin reload, while by /is missions work without reload. Additionaly, changing mission types between player/island requires a double plugin reload. When I previously storing categories in

cache the value in the MissionsContainer and expose a method of it to MissionsManager, it will probably fix the issue
Don't listen to SETTINGS_UPDATE, rather update it when a mission is registered/unregistered

Lemvs:
> > ing a category via menu without having an island only works after a plugin reload, while by /is missions work without reload. Additionaly, changing mission types between player/island requires a double plugin reload. When I previously storing categories in
> 
> cache the value in the MissionsContainer and expose a method of it to MissionsManager, it will probably fix the issue Don't listen to SETTINGS_UPDATE, rather update it when a mission is registered/unregistered

I wanted to do that initially, but you tell me to use SETTINGS_UPDATE_EVENT, so I preferred to ask 😅
Now everything works 🙂

src/main/java/com/bgsoftware/superiorskyblock/mission/container/DefaultMissionsContainer.java
```diff
@@ -8,13 +8,17 @@
...
+import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Set;
 import java.util.function.Predicate;
 
 public class DefaultMissionsContainer implements MissionsContainer {
 
+    private final Set<MissionCategory> PLAYER_MISSION_CATEGORIES_CACHE = new HashSet<>();
```
OmerBenGera: camelCase

src/main/java/com/bgsoftware/superiorskyblock/mission/container/MissionsContainer.java
```diff
@@ -32,6 +32,10 @@ public interface MissionsContainer {
 
     List<MissionCategory> getMissionCategories();
 
+    boolean isPlayerMissionCategory(MissionCategory missionCategory);
+
+    boolean isPlayerMissionCategories();
```
OmerBenGera: `hasAnyPlayerMissionCategories`

OmerBenGera (changes requested):

OmerBenGera (approved):

PR #2766, opened 2025-11-08
Fixed dimension usage in commands

Lemvs:
- Some commands, when they required a dimension, had "world" argument in their usage, others had "normal/nether/the_end" so to standardize this, I introduced the "dimension" argument.
- I also changed the getDimensions() method in the tab completes to include only the enabled dimension.
- I also changed the INVALID_ENVIRONMENT message to INVALID_DIMENSION, as the old one used the old name and inform about "Invalid world ..." instead of "Invalid dimension ..."


src/main/java/com/bgsoftware/superiorskyblock/core/messages/Message.java
```diff
@@ -430,9 +431,9 @@ public void send(CommandSender sender, Locale locale, Object... args) {
     INVALID_BIOME,
     INVALID_BLOCK,
     INVALID_BORDER_COLOR,
+    INVALID_DIMENSION,
     INVALID_EFFECT,
     INVALID_ENTITY,
-    INVALID_ENVIRONMENT,
```
OmerBenGera, line 435: Let's keep the INVALID_ENVIRONMENT, so people don't need to re-format the `INVALID_DIMENSION` which has the exact same purpose.

OmerBenGera (changes requested):

OmerBenGera (approved):

PR #2767, opened 2025-11-08
Improved disabling the upgrades module

Lemvs:
# Changelog # 
- I added an "enabled" option to the upgrades module config because if we disabled all options from this config, the module was disabled, which shouldn't be the case, as we might still want to use the upgrades menu to upgrade the team limit or island size, which are part of the main plugin, so still work despite these options being disabled. 
- I moved the block and entity limit modification commands to the upgrades module because you can specify whether these features are enabled in the module, but these commands didn't belong to the module, but to the main plugin, and they were always enabled.

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/module/upgrades/type/UpgradeTypeBlockLimits.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/module/upgrades/type/UpgradeTypeBlockLimits.java
@@ -14,6 +14,9 @@ import com.bgsoftware.superiorskyblock.core.key.ConstantKeys;
 import com.bgsoftware.superiorskyblock.core.key.Keys;
 import com.bgsoftware.superiorskyblock.core.messages.Message;
 import com.bgsoftware.superiorskyblock.core.mutable.MutableObject;
+import com.bgsoftware.superiorskyblock.module.upgrades.commands.CmdAdminAddBlockLimit;
+import com.bgsoftware.superiorskyblock.module.upgrades.commands.CmdAdminRemoveBlockLimit;
+import com.bgsoftware.superiorskyblock.module.upgrades.commands.CmdAdminSetBlockLimit;
 import com.bgsoftware.superiorskyblock.world.BukkitItems;
 import org.bukkit.Location;
 import org.bukkit.Material;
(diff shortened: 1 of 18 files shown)


src/main/java/com/bgsoftware/superiorskyblock/module/upgrades/type/UpgradeTypeBlockLimits.java
```diff
@@ -33,11 +36,15 @@
...
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
 public class UpgradeTypeBlockLimits implements IUpgradeType {
 
+    private static final List<ISuperiorCommand> commands = Arrays.asList(new CmdAdminAddBlockLimit(),
+            new CmdAdminRemoveBlockLimit(), new CmdAdminSetBlockLimit());
+
```
OmerBenGera: Redundant, can be created and returned by `getCommands`

src/main/java/com/bgsoftware/superiorskyblock/module/upgrades/type/UpgradeTypeEntityLimits.java
```diff
@@ -46,6 +49,9 @@
 
 public class UpgradeTypeEntityLimits implements IUpgradeType {
 
+    private static final List<ISuperiorCommand> commands = Arrays.asList(new CmdAdminAddEntityLimit(),
+            new CmdAdminRemoveEntityLimit(), new CmdAdminSetEntityLimit());
+
```
OmerBenGera: Redundant, can be created and returned by `getCommands`

src/main/resources/modules/upgrades/config.yml
```diff
@@ -5,6 +5,8 @@
 ##                                                  ##
 ######################################################
 
+enabled: true
```
OmerBenGera, line 10: Comment the field

OmerBenGera (changes requested):

src/main/resources/modules/upgrades/config.yml
Lemvs: In other modules, except for the mission, this option does not have a comment, so in other modules also add such a comment?

src/main/java/com/bgsoftware/superiorskyblock/module/upgrades/type/UpgradeTypeBlockLimits.java
Lemvs: In other types of upgrades it is done like this, so I did the same. So should I change it in other types too?

src/main/resources/modules/upgrades/config.yml
OmerBenGera: > In other modules, except for the mission, this option does not have a comment, so in other modules also add such a comment?

Yep

OmerBenGera (approved):

PR #2807, opened 2025-12-19
feat. new flags regarding beehives

nivcoo:
Hello,

I added a new flags regarding beehives spawn.

Because actually it's not a real "natural spawn" if bees are inside beehives.


src/main/java/com/bgsoftware/superiorskyblock/listener/IslandFlagsListener.java
```diff
@@ -82,6 +90,14 @@ private void onEntitySpawn(GameEvent<GameEventArgs.EntitySpawnEvent> e) {
     private boolean checkPreventEntitySpawn(GameEvent<GameEventArgs.EntitySpawnEvent> e, Location entityLocation) {
         CreatureSpawnEvent.SpawnReason spawnReason = e.getArgs().spawnReason;
 
+        if (BEE_ENTITY_TYPE != null &&
```
OmerBenGera: Technically no need to check for bee entity type, as checking for the beehive spawn reason is enough

src/main/java/com/bgsoftware/superiorskyblock/listener/IslandFlagsListener.java
```diff
@@ -82,6 +90,14 @@ private void onEntitySpawn(GameEvent<GameEventArgs.EntitySpawnEvent> e) {
...
+                e.getArgs().entity.getType() == BEE_ENTITY_TYPE &&
+                BEEHIVE_SPAWN_REASON != null &&
```
OmerBenGera: No need to check if it's null, only if `spawnReason == BEEHIVE_SPAWN_REASON` (spawnReason will never be null)

src/main/resources/menus/settings.yml
```diff
@@ -471,6 +471,25 @@ settings:
       lore:
         - '&7Allow withers to explode blocks inside your island.'
         - '&7Currently &cDISABLED&7.'
+    sound:
+      type: ORB_PICKUP
+      volume: 0.2
+      pitch: 0.2
+
+  beehive_bees_spawn:
```
OmerBenGera: Doesn't exist in this version

src/main/resources/menus/settings1_12.yml
```diff
@@ -474,4 +474,23 @@ settings:
     sound:
       type: ENTITY_EXPERIENCE_ORB_PICKUP
       volume: 0.2
-      pitch: 0.2
\ No newline at end of file
+      pitch: 0.2
+
+  beehive_bees_spawn:
```
OmerBenGera: Doesn't exist in this version

src/main/resources/menus/settings1_13.yml
```diff
@@ -457,4 +457,23 @@ settings:
     sound:
       type: ENTITY_EXPERIENCE_ORB_PICKUP
       volume: 0.2
-      pitch: 0.2
\ No newline at end of file
+      pitch: 0.2
+
+  beehive_bees_spawn:
```
OmerBenGera: Doesn't exist in this version

src/main/resources/menus/settings1_20.yml
```diff
@@ -465,4 +465,23 @@ settings:
     sound:
       type: ENTITY_EXPERIENCE_ORB_PICKUP
       volume: 0.2
+      pitch: 0.2
+
+  beehive_bees_spawn:
```
OmerBenGera: Should be sorted alphabetic

src/main/java/com/bgsoftware/superiorskyblock/island/flag/IslandFlags.java
```diff
@@ -30,6 +30,7 @@ public class IslandFlags {
     public static final IslandFlag TREE_GROWTH = register("TREE_GROWTH");
     public static final IslandFlag WATER_FLOW = register("WATER_FLOW");
     public static final IslandFlag WITHER_EXPLOSION = register("WITHER_EXPLOSION");
+    public static final IslandFlag BEEHIVE_BEES_SPAWN = register("BEEHIVE_BEES_SPAWN");
```
OmerBenGera: Should not be registered if bees are not in the game

OmerBenGera (changes requested):

nivcoo:
Ok i fixed everything you wanted except the last one : "Should not be registered if bees are not in the game"

How i've to do that ? and I add a new file settings1_15.yml does it will detect it automaticaly ? I don't see any reference in code

Lemvs:
There is no need to add the file for 1.15, you could have renamed the file for 1.13, but that wasn't necessary either, because the closest supported versions are 1.12 and 1.16, nothing in between is supported. After your current changes, the file for 1.13 is completely unnecessary, no one version will use it.

Lemvs:
And in the IslandFlags class, mark this flag as Nullable and it would also be better to arrange it alphabetically 😉

nivcoo:
Yes of course I forget to order that in code too. And so what I have to do for the config file ? Remove the 1.13 one or just add the flags into the 1.13 file ?

Lemvs:
It doesn't really matter, but since the flag requires version 1.15, I guess you could create a file for 1.15 and remove the one for 1.13, but maybe it's better to let Ome say what he prefers

nivcoo:
Or maybe just create a 1.16 file because the 1.15 is not supported

Lemvs:
Right, you can do it that way. The Permissions menu is also for 1.16, and the previous are for 1.9, so it should be okay

nivcoo:
I will do the next commit in that way and fix nullable and order in code and it should be good

src/main/java/com/bgsoftware/superiorskyblock/island/flag/IslandFlags.java
```diff
@@ -1,10 +1,15 @@
 package com.bgsoftware.superiorskyblock.island.flag;
 
+import com.bgsoftware.common.annotations.NotNull;
+import com.bgsoftware.common.annotations.Nullable;
 import com.bgsoftware.superiorskyblock.api.island.IslandFlag;
+import com.bgsoftware.superiorskyblock.api.island.IslandPrivilege;
```
OmerBenGera: No need to import it

src/main/java/com/bgsoftware/superiorskyblock/listener/IslandFlagsListener.java
```diff
@@ -82,6 +86,10 @@ private void onEntitySpawn(GameEvent<GameEventArgs.EntitySpawnEvent> e) {
     private boolean checkPreventEntitySpawn(GameEvent<GameEventArgs.EntitySpawnEvent> e, Location entityLocation) {
         CreatureSpawnEvent.SpawnReason spawnReason = e.getArgs().spawnReason;
 
+        if (spawnReason == BEEHIVE_SPAWN_REASON) {
+            return preventAction(entityLocation, IslandFlags.BEEHIVE_BEES_SPAWN);
+        }
+
```
OmerBenGera: Move this down below to the `if-else if` clause

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/listener/IslandFlagsListener.java
nivcoo: good

OmerBenGera:
Can be done with entity categories - closing this

PR #2857, opened 2026-02-04
Added the ability to toggle when a player should receive help

Lemvs:
We can now change whether players will see help or a special message when using a command that doesn't exist or for which they don't have permission. This also applies to the admin command, as it previously sent an invalid amount message, even if we entered a command, not an amount.


src/main/java/com/bgsoftware/superiorskyblock/commands/CommandsManagerImpl.java
```diff
@@ -305,39 +309,33 @@ public boolean execute(CommandSender sender, String label, String[] args) {
...
                         subCommandToExecute = "panel";
-                    } else {
+                    else
                         subCommandToExecute = "tp";
-                    }
-
-                    // We don't want to end up in an infinite loop
-                    if (!subCommandToExecute.equalsIgnoreCase(executedSubCommand)) {
-                        dispatchSubCommand(sender, subCommandToExecute);
-                    }
```
OmerBenGera, line 330: Why did you remove this?

src/main/java/com/bgsoftware/superiorskyblock/commands/CommandsManagerImpl.java
```diff
@@ -305,39 +309,33 @@ public boolean execute(CommandSender sender, String label, String[] args) {
...
+                    dispatchSubCommand(sender, subCommandToExecute);
                     return false;
                 }
             }
 
-            // We don't want to end up in an infinite loop
-            if (!"help".equalsIgnoreCase(executedSubCommand)) {
-                dispatchSubCommand(sender, "help");
-            }
-
```
OmerBenGera, line 340: Why did you remove this?

OmerBenGera (changes requested):
Do not have if-lines without curly brackets

OmerBenGera:
I gave you a CR. Fix the issues and I'll look into it.

Lemvs:
> Do not have if-lines without curly brackets

Why? Even a few lines down in the admin command, you're using it that way 🤔

image


And for

> Why did you remove this?

I have already written why

OmerBenGera (approved):

PR #2860, opened 2026-02-05
Added whitelists and blacklists for mob drops multiplier

Lemvs:
As suggested in this [discussion](https://github.com/BG-Software-LLC/SuperiorSkyblock2/discussions/149), but I also added whitelist and blacklist for entities, not just for items


src/main/java/com/bgsoftware/superiorskyblock/module/upgrades/type/UpgradeTypeMobDrops.java
```diff
@@ -105,6 +119,9 @@ private void modifyEventDrops(List<ItemStack> drops, LivingEntity livingEntity,
 
             for (ItemStack itemStack : drops) {
                 if (itemStack != null && !BukkitEntities.isEquipment(livingEntity, itemStack)) {
+                    if (!isWhitelisted(itemStack) || isBlacklisted(itemStack))
```
OmerBenGera, line 114: You run Key#of twice here, can be optimized

src/main/java/com/bgsoftware/superiorskyblock/module/upgrades/type/UpgradeTypeMobDrops.java
```diff
@@ -79,14 +90,17 @@ public void onEntityDeath(EntityDeathEvent e) {
             if (!canDupeDropsForEntity(e.getEntity()))
                 return;
 
+            if (!isWhitelisted(e.getEntity()) || isBlacklisted(e.getEntity()))
```
OmerBenGera, line 85: You run Key#of twice here, can be optimized

OmerBenGera (changes requested):

Lemvs:
@OmerBenGera I understand that you don't have much time to introduce something yourself, but if even pull requests have to wait a few months for re-verification, it will soon die completely... 🙃😞

src/main/java/com/bgsoftware/superiorskyblock/module/upgrades/type/UpgradeTypeMobDrops.java
```diff
@@ -25,15 +26,25 @@
...
+    private final Set<Key> whitelistedEntities;
+    private final Set<Key> blacklistedEntities;
+
     public UpgradeTypeMobDrops(SuperiorSkyblockPlugin plugin) {
         this.plugin = plugin;
         isWildStackerInstalled = Bukkit.getPluginManager().isPluginEnabled("WildStacker");
+        whitelistedItems = plugin.getSettings().getMobDrops().getWhitelistedItems();
+        blacklistedItems = plugin.getSettings().getMobDrops().getBlacklistedItems();
+        whitelistedEntities = plugin.getSettings().getMobDrops().getWhitelistedEntities();
+        blacklistedEntities = plugin.getSettings().getMobDrops().getBlacklistedEntities();
```
OmerBenGera: Do not cache this value here, always access the SettingsManager

OmerBenGera (changes requested):

OmerBenGera (changes requested):
Fix conflicts

API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
```diff
@@ -1452,6 +1480,27 @@ interface MobDrops {
...
+
+        /**
+         * Get an entity category by its name
+         *
+         * @param name The name of the category.
+         */
+        @Nullable
+        EntityCategory getCategoryByName(String name);
 
     }
```
OmerBenGera, line 1478: Something is weird with indents

OmerBenGera (changes requested):

API/src/main/java/com/bgsoftware/superiorskyblock/api/config/SettingsManager.java
Lemvs: Uhm, I just accepted both changes

Lemvs: If you go to all commits of this pr, instead of just the merging one, this code is not modified, so it looks okay

Lemvs:
@OmerBenGera what about this? I did everything you asked for 😕

Lemvs:
But now I'm wondering if it wouldn't be better to move that section from the main config to the upgrades module config; since that's where the feature can be enabled, that's also where it should be configurable. The same applies to the options regarding the crop growth multiplier.

OmerBenGera:
Yeah, should be in the upgrades module configuration

Lemvs:
I'll check if everything is working tomorrow.

Lemvs:
> Yeah, should be in the upgrades module configuration

Everything should work correctly now, you can review the code.

Lemvs:
Probably the same problem with convert as in https://github.com/BG-Software-LLC/SuperiorSkyblock2/pull/3003

NMS/src/main/templates/crops/CropsTickingMethod.java.template
```diff
@@ -54,14 +53,16 @@ public abstract class CropsTickingMethod {
...
         CROPS_TO_GROW_CACHE = new HashSet<>();
-        plugin.getSettings().getCropsToGrow().forEach(cropName -> {
-            Key key = Keys.ofMaterialAndData(cropName);
-            if (key instanceof MaterialKey materialKey) {
-                Block block = CraftMagicNumbers.getBlock(materialKey.getMaterial());
-                if (block != null && block.defaultBlockState().isRandomlyTicking())
-                    CROPS_TO_GROW_CACHE.add(block);
-            }
-        });
+        PluginModule pluginModule = plugin.getModules().getModule("upgrades");
```
OmerBenGera, line 56: Why not using BuiltinModules.UPGRADES?

OmerBenGera (changes requested):

OmerBenGera:
Great work.
I think I will wait with merging it to the next version, however let's get this PR approved until then :)
Only 1 change is required, other than that looks good.

Lemvs:
> Probably the same problem with convert as in https://github.com/BG-Software-LLC/SuperiorSkyblock2/pull/3003

This still needs to be solved, because I think I tested convert here before I deleted these options from the original config

Lemvs:
# Changelog #
- Moved the configuration converters for the Bank, Generators, and Upgrades modules from their respective files to `SettingsManagerImpl`.
- Changed a bit the configuration converters code within `SettingsManagerImpl`, to match the same style.
- Moved the order of world configuration conversion - the latest changes regarding custom dimension creation should be processed at the very end of the convertData() method, since the method involves multiple other conversion steps for worlds section, those must execute first to avoid data loss.
- Changed the default values ​​for bank-worth-rate and disband-refund in the bank module configuration to ensure they are treated as doubles (i.e., including a .0 decimal).

PR #2874, opened 2026-02-22
Fixes Players Without Islands Treated as Same Team (Pvp)

### LINKED ISSUE #2873: Players Without Islands Treated as Same Team (Pvp)
### Minecraft's Version

This server is running Paper version 1.21.1-133-ver/1.21.1@3cb8529 (2025-03-28T16:14:36Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)
You are running the latest version
Previous version: 1.21.1-119-7cd4f2c (MC: 1.21.1)

### Plugin's Version

SuperiorSkyblock2 version 2025.2.1
Feature packed Skyblock core.
Website: https://bg-software.com/
Author: Ome_R

### Describe the bug

Players without islands are currently being treated as if they are on the same team (PvP).

`SuperiorPlayer.canHit(SuperiorPlayer otherPlayer)` checks whether both players belong to the same island. However, if neither player is in an island, the method ends up comparing two `null` island values. Since `null == null` evaluates to `true`, the method incorrectly treats them as being on the same island.

### To Reproduce

1. Install SuperiorSkyblock2.
2. Disable friendly fire in the configuration.
3. Join the server with two separate accounts without creating an island on either account.
4. Attempt to attack one account with the other.
5. The attack is cancelled and `Message.HIT_ISLAND_MEMBER` is sent, even though neither player belongs to an island.

### Additional Information

_No response_

CdeJong:
Fixed an issue where players without islands were being treated as if they were on the same team (PvP).

Fixes #2873

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/player/SSuperiorPlayer.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/player/SSuperiorPlayer.java
@@ -407,7 +407,7 @@ public class SSuperiorPlayer implements SuperiorPlayer {
         World world = getWorld();
 
         // Checks for island teammates pvp
-        if (getIsland() == otherPlayer.getIsland() && (world == null || !isPvPWorldInternal(world.getName())))
+        if (getIslandLeader().equals(otherPlayer.getIslandLeader()) && (world == null || !isPvPWorldInternal(world.getName())))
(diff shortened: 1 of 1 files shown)


OmerBenGera:
Better solution would be to check if island is not null and the islands are equal

OmerBenGera:
Fixed in another PR, thanks for the contribution!

PR #2895, opened 2026-03-29
Added warning about missing permissions and settings in their menu files

Lemvs:
Just like there is a warning about missing blocks in interactables.yml, I also added a warning about missing permissions in menus/permissions.yml and settings in menus/settings.yml. New permissions appear from time to time and if someone is not up to date with changes, they will not know that they are missing some in the file.

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/core/menu/impl/MenuIslandPrivileges.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/core/menu/impl/MenuIslandPrivileges.java
@@ -13,6 +13,7 @@ import com.bgsoftware.superiorskyblock.api.wrappers.SuperiorPlayer;
 import com.bgsoftware.superiorskyblock.core.LazyReference;
 import com.bgsoftware.superiorskyblock.core.io.MenuParserImpl;
 import com.bgsoftware.superiorskyblock.core.itemstack.ItemBuilder;
+import com.bgsoftware.superiorskyblock.core.logging.Log;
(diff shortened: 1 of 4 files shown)


OmerBenGera:
The issue with that is that in this time, not all island flags and privileges are loaded yet. They can be loaded afterwards

Lemvs:
> The issue with that is that in this time, not all island flags and privileges are loaded yet. They can be loaded afterwards

I know, but it was meant to inform about missing built-in permissions and flags, I can't deal with the fact that permissions and flags registered by other plugins are not known here yet 🫤

src/main/java/com/bgsoftware/superiorskyblock/core/menu/impl/MenuIslandPrivileges.java
```diff
@@ -91,17 +96,23 @@ public static MenuIslandPrivileges createInstance() {
         String higherRolePermission = cfg.getString("messages.higher-role-permission", "");
 
         List<MenuIslandPrivileges.IslandPrivilegeInfo> islandPrivileges = new LinkedList<>();
+        Set<String> missingPrivileges = new HashSet<>(IslandPrivileges.getPrivilegesNamesAsSet());
```
OmerBenGera: Make this a set of strings of the detected privileges, then below iterate through IslandPrivilege#values and check if it doesn't exist in the set. Maybe "less optimized" but more readable

OmerBenGera (changes requested):

OmerBenGera:
Fix conflicts

OmerBenGera (approved):

PR #2910, opened 2026-04-20
Added support for setting custom biomes for islands and island worlds

Lemvs:
I've added the ability to set a custom biomes created using datapacks for islands and island worlds. This change required adding support for NamespacedKey, so instead of using Biome.valueOf(), I've created a new method in the NMS Adapter that allows you to retrieve a biome with a specified namespace in new versions.

So now you can specify the biome in one of these three ways, it is not case sensitive, it will be adjusted depending on the version:
- Without namespace:
  `biome: PLAINS`
- With minecraft namespace (1.19+):
  `biome: minecraft:plains`
- With custom namespace (1.19+):
  `biome: my_datapack:my_biome`
  
When you merge, close this discussion: https://github.com/BG-Software-LLC/SuperiorSkyblock2/discussions/2908 😊

I've also improved the readability of the setTrim method, adding it for NMS 1_20_3 and 1_20_4 (I don't know why I skipped those versions before) and removing it from NMS 1_19, as it was experimental there and couldn't be used without a datapack. Currently, it just displays a bunch of warnings in that file.


OmerBenGera (approved):

OmerBenGera:
I was just working on this exact same feature, but I did it way more complicated than your solution. Good job :)

PR #2911, opened 2026-04-21
Added /is localchat command

Lemvs:
I added the /is localchat [message] command, which allows you to send a message to players who are inside the same island. Originally, the command was supposed to be called /is chat, but due to existing classes/methods for team chat that used just "Chat" instead of "TeamChat" in their names, I found an alternate name. However, there's nothing stopping you from adding this alias to this command using the command-aliases option in config. 😊

To do this:
- I deprecated the IslandChatEvent event because its name had become too generic and could cause confusion. Instead, I created IslandTeamChatEvent, which extends IslandChatEvent for backward compatibility. Once some time has passed and the old event can be removed, its methods will simply be moved to the new one.
- I added new events, IslandLocalChatEvent and PlayerToggleLocalChatEvent, which allow you to track actions related to local chat.
- I added the IslandUtils.sendMessageToAllPlayersInside() method and I changed the IslandUtils.sendMessage() method in IslandUtils.sendMessageToMembers() for consistency.
- I added the IslandUtils.handleIslandLocalChat() method, based on the handleIslandTeamChat() method, which was previously called IslandUtils.handleIslandChat().
- I added the CmdLocalChat command based on the CmdTeamChat command, and created corresponding messages in a similar way.
- In both commands, if we want to enable one chat while the other is also enabled, the other one will be disabled to avoid writing on two chats in one time.
- I changed the CommandArguments.getIslandWhereStanding() method in CommandArguments.getIslandWhereStandingOrSenderIsland(), and added the CommandArguments.getIslandWhereStanding() method, which actually returns the island we are standing on, because it introduced this is confusing.

These are probably the most important changes. If I've made any mistakes or if there's an easier way, let me know. 😉

After merging, you can close this discussion: https://github.com/BG-Software-LLC/SuperiorSkyblock2/discussions/2379


API/src/main/java/com/bgsoftware/superiorskyblock/api/wrappers/SuperiorPlayer.java
```diff
@@ -443,6 +443,23 @@ public interface SuperiorPlayer extends IMissionsHolder, IPersistentDataHolder,
...
+     */
+    void toggleLocalChat();
+
+    /**
+     * Set whether the local chat is enabled for the player.
+     *
+     * @param enabled true to enable local chat.
+     */
+    void setLocalChat(boolean enabled);
+
```
OmerBenGera: Let's add a "ChatState" that can be registered by the API. You will initialize 3 default ones - "TEAM_CHAT", "LOCAL_CHAT" and "GLOBAL". It will have a method that takes a player and returns the target players that needs to get the message, or null if all of the players should get it. Then have only one method to set the chat state, and deprecate the rest of the methods.

API/src/main/java/com/bgsoftware/superiorskyblock/api/events/IslandChatEvent.java
```diff
@@ -6,8 +6,11 @@
 import org.bukkit.event.Cancellable;
 
 /**
- * IslandChatEvent is called when a player talks in islands chat.
+ * IslandChatEvent is called when a player talks in island team chat.
+ * @deprecated This event has been renamed to {@link IslandTeamChatEvent}
+ * and will be removed in the future.
  */
+@Deprecated
```
OmerBenGera: Do not deprecate it. Instead, it should be called for local chat as well

API/src/main/java/com/bgsoftware/superiorskyblock/api/events/PlayerToggleLocalChatEvent.java
```diff
@@ -0,0 +1,56 @@
...
+import com.bgsoftware.superiorskyblock.api.wrappers.SuperiorPlayer;
+import org.bukkit.Bukkit;
+import org.bukkit.event.Cancellable;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+
+/**
+ * PlayerToggleLocalChatEvent is called when a player toggles his local chat.
+ */
+public class PlayerToggleLocalChatEvent extends Event implements Cancellable {
```
OmerBenGera: Change it to PlayerChatStateChangeEvent or something like that

OmerBenGera (changes requested):

API/src/main/java/com/bgsoftware/superiorskyblock/api/wrappers/SuperiorPlayer.java
Lemvs: I was thinking about it because I was also thinking about whether to separate the admin spy option for these chats, but I was thinking more about the name "ChatChannel", what do you prefer?

And then maybe method to get players who should spy that messages? But where store it? Because I think always getting Bukkit.getOnlinePlayers() is not needed, rather few players has spy enabled.

API/src/main/java/com/bgsoftware/superiorskyblock/api/events/PlayerToggleLocalChatEvent.java
Lemvs: But I could simply refactor this class or create a new one, but then what about this one? Because I probably shouldn't just mess with something that's already in the API.

Lemvs:
@OmerBenGera I did something like this (for now I only didn't change the ChatState toggle event). Let me know if this is what you had in mind and what could be improved.

API/src/main/java/com/bgsoftware/superiorskyblock/api/events/IslandChatEvent.java
```diff
@@ -1,33 +1,47 @@
...
+     * @param island         The island from which the message was sent.
+     * @param superiorPlayer The player who sent the message.
+     * @param message        The message that was sent.
+     * @param chatState      The ChatState in which the message was sent.
+     */
+    public IslandChatEvent(Island island, SuperiorPlayer superiorPlayer, String message, ChatState chatState) {
         super(island);
         this.superiorPlayer = superiorPlayer;
         this.message = message;
+        this.chatState = chatState;
```
OmerBenGera: This is not necessary as developers could use SuperiorPlayer#getChatState

src/main/java/com/bgsoftware/superiorskyblock/player/SSuperiorPlayer.java
```diff
@@ -703,6 +708,37 @@ public void setBlocksStacker(boolean enabled) {
...
 
+    @Override
+    public ChatState getChatState() {
+        return this.chatState;
+    }
+
+    @Override
+    public boolean hasChatState(ChatState chatState) {
+        return this.chatState == chatState;
+    }
```
OmerBenGera: Can be removed

src/main/java/com/bgsoftware/superiorskyblock/player/SSuperiorPlayer.java
```diff
@@ -703,6 +708,37 @@ public void setBlocksStacker(boolean enabled) {
...
+    @Override
+    public void setChatState(ChatState chatState) {
+        Log.debug(Debug.SET_CHAT_STATE, getName(), chatState);
+        this.chatState = chatState;
+    }
+
+    @Override
+    public void addSpiedChatState(ChatState chatState) {
+        this.spiedChatStates.add(chatState);
+    }
```
OmerBenGera: Spy chat should spy *all* chats, so this can be removed

API/src/main/java/com/bgsoftware/superiorskyblock/api/events/PlayerToggleLocalChatEvent.java
OmerBenGera: I'd deprecate `PlayerToggleTeamChatEvent` and have only one single event for switching chat states

src/main/java/com/bgsoftware/superiorskyblock/core/events/args/PluginEventArgs.java
```diff
@@ -200,6 +201,12 @@ public static class IslandChangeRoleLimit extends IslandDoActionArgs {
 
     }
 
+    public static class IslandChangeRolePrivilege extends IslandDoActionArgs {
+
+        public PlayerRole playerRole;
+
+    }
+
```
OmerBenGera, line 209: ?

src/main/java/com/bgsoftware/superiorskyblock/core/events/plugin/PluginEventType.java
```diff
@@ -507,6 +509,12 @@ public void applyBukkitToPluginEvent(Event bukkitEvent, PluginEvent<IslandChange
             pluginEvent.getArgs().roleLimit = ((IslandChangeRoleLimitEvent) bukkitEvent).getRoleLimit();
         }
     };
+    public static final PluginEventType<IslandChangeRolePrivilege> ISLAND_CHANGE_ROLE_PRIVILEGE_EVENT = new PluginEventType<IslandChangeRolePrivilege>(IslandChangeRolePrivilegeEvent.class) {
+        @Override
+        public Event createBukkitEvent(IslandChangeRolePrivilege args) {
+            return new IslandChangeRolePrivilegeEvent(args.island, args.superiorPlayer, args.playerRole);
+        }
+    };
```
OmerBenGera, line 517: ?

src/main/java/com/bgsoftware/superiorskyblock/island/IslandUtils.java
```diff
@@ -341,23 +354,41 @@ public static List<Biome> getDefaultWorldBiomes() {
...
+            spyChatFormat = Message.SPY_TEAM_CHAT_FORMAT;
+        }
+
+        String playerRoleName = getPlayerRole(island, superiorPlayer).getDisplayName();
+        String superiorPlayerName = superiorPlayer.getName();
+        String formattedMessage = event.getArgs().message;
 
-        Message.SPY_TEAM_CHAT_FORMAT.send(Bukkit.getConsoleSender(), superiorPlayer.getPlayerRole().getDisplayName(),
-                superiorPlayer.getName(), event.getArgs().message);
+        for (SuperiorPlayer targetPlayer : chatState.getTargetPlayers(superiorPlayer)) {
```
OmerBenGera: Because you don't support ChatState, the `getTargetPlayers` is redundant now and can be removed.

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/core/events/args/PluginEventArgs.java
Lemvs: I just put it in the right order because the rest are alphabetical and this isn't

src/main/java/com/bgsoftware/superiorskyblock/core/events/plugin/PluginEventType.java
Lemvs: I just put it in the right order because the rest are alphabetical and this isn't

src/main/java/com/bgsoftware/superiorskyblock/island/IslandUtils.java
Lemvs: So why would I have to do such a method and allow external ChatState registering, I don't understand...

API/src/main/java/com/bgsoftware/superiorskyblock/api/events/PlayerToggleLocalChatEvent.java
Lemvs: But since it will be deprecated, it should still work until it is removed, so I have to call two events?

API/src/main/java/com/bgsoftware/superiorskyblock/api/wrappers/SuperiorPlayer.java
Lemvs: @OmerBenGera ^ (second part)

Lemvs:
Sorry for the duplicate commits, but I had to recreate the project because the missions you recently moved to the main project wouldn't load correctly for me.

API/src/main/java/com/bgsoftware/superiorskyblock/api/events/PlayerToggleLocalChatEvent.java
OmerBenGera: Yeah

src/main/java/com/bgsoftware/superiorskyblock/island/IslandUtils.java
OmerBenGera: So it can be manipulated in the IslandChatEvent. I thought at first the plugin will handle sending messages even for custom ChatState, but its more conplicated to support it and developers could just do it themselves using IslandChatEvent

Lemvs: It makes sense at all to be able to register it, isn't it better to make an enum out of it if it's not supposed to be somehow expanded?

Lemvs: @OmerBenGera tell me what to do, because I don't know how to solve it anymore ^ 🫤

OmerBenGera: > It makes sense at all to be able to register it, isn't it better to make an enum out of it if it's not supposed to be somehow expanded?

No, ENUMs should not be used in the API and Enumerable is the correct way.

src/main/java/com/bgsoftware/superiorskyblock/commands/player/CmdTeamChat.java
```diff
@@ -64,16 +66,26 @@ public void execute(SuperiorSkyblockPlugin plugin, CommandSender sender, String[
         SuperiorPlayer superiorPlayer = arguments.getSuperiorPlayer();
 
         if (args.length == 1) {
-            if (!PluginEventsFactory.callPlayerToggleTeamChatEvent(superiorPlayer))
+            ChatState oldChatState = superiorPlayer.getChatState();
+            ChatState newChatState = oldChatState == ChatStates.TEAM_CHAT ?
+                    ChatStates.GLOBAL : ChatStates.TEAM_CHAT;
+
+            //TODO What to do with it? Call two events, because the second one is deprecated, so should still works?
```
OmerBenGera: Yes, you should call both events - exactly as you did.

src/main/java/com/bgsoftware/superiorskyblock/island/IslandUtils.java
```diff
@@ -341,23 +354,41 @@ public static List<Biome> getDefaultWorldBiomes() {
...
-                Message.SPY_TEAM_CHAT_FORMAT.send(onlinePlayer, superiorPlayer.getPlayerRole().getDisplayName(),
-                        superiorPlayer.getName(), event.getArgs().message);
+        spyChatFormat.send(Bukkit.getConsoleSender(), playerRoleName, superiorPlayerName, formattedMessage);
+
+        //TODO Collect somewhere spies instead of using Bukkit.getOnlinePlayers() always...
+        for (Player player : Bukkit.getOnlinePlayers()) {
+            SuperiorPlayer onlinePlayer = plugin.getPlayers().getSuperiorPlayer(player);
+            if (onlinePlayer.hasAdminSpyEnabled()) {
+                spyChatFormat.send(onlinePlayer, playerRoleName, superiorPlayerName, formattedMessage);
+            }
```
OmerBenGera: Let's create a new class called "IslandChat" - it will have the handleIslandChat method from IslandUtils, and also will track all chat spy.

src/main/java/com/bgsoftware/superiorskyblock/listener/PlayersListener.java
```diff
@@ -395,24 +396,43 @@ private void onPlayerChatLowest(GameEvent<GameEventArgs.PlayerChatEvent> e) {
...
 
             e.setCancelled();
 
-            String message = e.getArgs().message;
+            IslandUtils.handleIslandChat(island, superiorPlayer, e.getArgs().message);
+        } else if (superiorPlayer.getChatState() == ChatStates.TEAM_CHAT) {
+            Island island = superiorPlayer.getIsland();
+
+            if (island == null) {
+                //TODO What to do with it? Call two events, because the second one is deprecated, so should still works?
```
OmerBenGera: You did good.

API/src/main/java/com/bgsoftware/superiorskyblock/api/player/chat/ChatState.java
```diff
@@ -0,0 +1,88 @@
...
+    public final String getName() {
+        return name;
+    }
+
+    /**
+     * Get the target players for the player.
+     *
+     * @param superiorPlayer The player to get the target players for.
+     */
+    public abstract List<SuperiorPlayer> getTargetPlayers(SuperiorPlayer superiorPlayer);
```
OmerBenGera: Remove this.
Once you did it, you can change the class to be regular and not abstract

OmerBenGera (changes requested):

Lemvs:
@OmerBenGera I don't know if this is what you meant by the IslandChat class, if not, I need more information.

API/src/main/java/com/bgsoftware/superiorskyblock/api/player/chat/ChatState.java
OmerBenGera: Should be private, similar to all regular Enumerables.

OmerBenGera: Should take name as an argument and create the ChatState on its own.

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/island/IslandChat.java
```diff
@@ -0,0 +1,80 @@
...
+import com.bgsoftware.superiorskyblock.core.messages.Message;
+import com.bgsoftware.superiorskyblock.player.chat.ChatStates;
+import org.bukkit.Bukkit;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class IslandChat {
+
+    private static final Set<SuperiorPlayer> spies = new HashSet<>();
```
OmerBenGera:
```suggestion
    private static final Set<SuperiorPlayer> spies = new LinkedHashSet<>();
```

OmerBenGera (changes requested):

OmerBenGera (approved):

PR #2916, opened 2026-04-24
Added an option to change the island biome in a given dimension

Lemvs:
I've added the ability to set the island's biome not only in the default Dimension. To achieve this:
- I've converted the Biomes menu from regular to paged, so that depending on the selected Dimension, it displays the appropriate Biomes. An additional advantage is the ability to create more than one page of biomes – new Biomes are constantly being created in the game, and you can also create your own Biomes using datapacks. While slots in the menu was previously limited, this won't be a problem now. The old menu format is, of course, converted accordingly for backward compatibility.
- I've added new methods: #getBiome(Dimension), #setBiome(Biome, Dimension), #setBiome(Biome, Dimension, boolean) to the Island object, replacing the old methods and marking them as Deprecated. The island now also stores a map created from the Dimension and Biomes, instead of a single Biome. 
- I've added ability to provide Dimension in the /is admin setbiome and /is biome command arguments; for the player command, this option can be disabled in the menu file.
- I've added ability to get the Dimension from the player's location in the /is biome command; this option can also be disabled in the menu file.
- I've modified the CHANGED_BIOME... messages to also display the Dimension in which the Biome was changed.
- I've modified the default biome menu configuration files, changed the item layout, added som of missing basic biomes, and demonstrated how to use the new feature in the version 1.16 file, adding Biomes for the Nether and The End.

There's just one small problem: after converting to the new format, the file is saved correctly. However, the biomes don't load, but other items load normally, and reloading the plugin fixes it. I don't know if I'm doing something wrong or if this is how MenuConverter works. I have no way to verify this because the old guis have been gone for a long time. I don't know if they're available in any version on Jenkins, probably not.

Additionally, I don't know if the way the dimension-selection-mode option works suits you. Maybe creating an enum instead of checking strings would be better. Let me know. If you have any other concerns, please let me know, just don't drag this out indefinitely, like responding to other Pull Requests... 🥹


API/src/main/java/com/bgsoftware/superiorskyblock/api/events/IslandBiomeChangeEvent.java
```diff
@@ -12,19 +14,34 @@ public class IslandBiomeChangeEvent extends IslandEvent implements Cancellable {
...
         this.biome = biome;
+        this.dimension = dimension;
+    }
+
+    /**
+     * The constructor for the event.
+     *
+     * @param superiorPlayer The player who changed the biome of the island.
+     * @param island         The island object that was changed.
+     * @param biome          The name of the new biome.
```
OmerBenGera, line 40: Add deprecated comment

API/src/main/java/com/bgsoftware/superiorskyblock/api/island/Island.java
```diff
@@ -1245,21 +1245,49 @@ List<CompletableFuture<Chunk>> getAllChunksAsync(Dimension dimension, @IslandChu
...
+    void setBiome(Biome biome, Dimension dimension);
+
+    /**
+     * Change the island biome in the dimension.
+     *
+     * @param biome        New biome to set.
+     * @param dimension    The dimension to change the biome in.
+     * @param updateBlocks Whether the blocks get updated or not.
+     */
+    void setBiome(Biome biome, Dimension dimension, boolean updateBlocks);
```
OmerBenGera: Change the boolean into an int flags, similar to IslandBlockFlags
Also, Dimension should probably be the first argument (idk why but in my head it's better)

src/main/java/com/bgsoftware/superiorskyblock/commands/player/CmdBiome.java
```diff
@@ -26,7 +33,11 @@ public String getPermission() {
 
     @Override
     public String getUsage(java.util.Locale locale) {
-        return "biome";
+        if (Menus.MENU_BIOMES.getDimensionSelectionMode().equals("ARGUMENT") ||
+                Menus.MENU_BIOMES.getDimensionSelectionMode().equals("BOTH"))
```
OmerBenGera: "DimensionSelectionMode" should be an enum, then compare enums and not strings

API/src/main/java/com/bgsoftware/superiorskyblock/api/hooks/MenusProvider.java
```diff
@@ -45,9 +46,22 @@ public interface MenusProvider {
      * @param targetPlayer The player to open the menu for.
      * @param previousMenu The previous menu that was opened, if exists.
      * @param targetIsland The island to change biomes for.
+     * @deprecated See {@link #openBiomes(SuperiorPlayer, ISuperiorMenu, Island, Dimension)}
      */
+    @Deprecated
     void openBiomes(SuperiorPlayer targetPlayer, @Nullable ISuperiorMenu previousMenu, Island targetIsland);
```
OmerBenGera: Should have a default implementation that throws UnsupportedOperationException.

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
```diff
@@ -246,7 +246,7 @@ protected IslandCache create() {
     private final AtomicReference<BigDecimal> bonusLevel = new AtomicReference<>(BigDecimal.ZERO);
     private final Map<MissionReference, Counter> completedMissions = new ConcurrentHashMap<>();
     private final Synchronized<IslandChest[]> islandChests = Synchronized.of(createDefaultIslandChests());
-    private final Synchronized<CompletableFuture<Biome>> biomeGetterTask = Synchronized.of(null);
+    private final Synchronized<Map<Dimension, CompletableFuture<Biome>>> biomeGetterTasks = Synchronized.of(new HashMap<>());
```
OmerBenGera: Use EnumerateMap instead.

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
```diff
@@ -274,7 +274,7 @@ protected IslandCache create() {
     private volatile String formattedName;
     private volatile String strippedName;
     private volatile String description;
-    private volatile Biome biome = null;
+    private final Map<Dimension, Biome> biomes = new ConcurrentHashMap<>();
```
OmerBenGera: Use EnumerateMap instead (Synchronized)

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
```diff
@@ -1976,42 +1976,58 @@ public void setPaypal(String paypal) {
 
     @Override
     public Biome getBiome() {
+        return getBiome(plugin.getSettings().getWorlds().getDefaultWorldDimension());
+    }
+
+    @Override
+    public Biome getBiome(Dimension dimension) {
+        Biome biome = biomes.get(dimension);
```
OmerBenGera: Make sure `dimension` is not null

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
```diff
@@ -1976,42 +1976,58 @@ public void setPaypal(String paypal) {
...
+
         if (biome == null) {
-            biomeGetterTask.set(this::getBiomeAsyncTask);
-            return IslandUtils.getDefaultWorldBiome(plugin.getSettings().getWorlds().getDefaultWorldDimension());
+            biomeGetterTasks.set(tasksMap -> {
```
OmerBenGera: `biomeGetterTasks.write`

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
```diff
@@ -1976,42 +1976,58 @@ public void setPaypal(String paypal) {
...
-                        if (error != null)
+                        if (error != null) {
                             newTask.completeExceptionally(error);
-                        else {
-                            this.biome = biome;
-                            biomeGetterTask.set((CompletableFuture<Biome>) null);
+                        } else {
+                            this.biomes.put(dimension, biome);
+
+                            biomeGetterTasks.set(tasksMap -> {
```
OmerBenGera: `biomeGetterTasks.write`

src/main/java/com/bgsoftware/superiorskyblock/listener/ChunksListener.java
```diff
@@ -115,17 +115,17 @@ private void handleChunkLoad(Chunk chunk, boolean isNewChunk) {
 
     private void handleIslandChunkLoad(Island island, Chunk chunk, ChunkPosition chunkPosition, boolean isNewChunk) {
         World world = chunk.getWorld();
-        Dimension dimension = plugin.getGrid().getIslandsWorldDimension(world);
+        Dimension dimension = WorldInfo.of(world).getDimension();
```
OmerBenGera: use `plugin.getGrid().getIslandsWorldDimension(world);`

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
```diff
@@ -2028,36 +2044,56 @@ public void setBiome(Biome biome) {
...
+        this.biomes.put(dimension, biome);
 
-        if (!updateBlocks)
+        if (!updateBlocks) {
             return;
+        }
 
-
-        IslandWorlds.accessIslandWorldsAsync(this, false, result -> {
+        IslandWorlds.accessIslandWorldAsync(this, dimension, false, result -> {
```
OmerBenGera: You cannot change the behavior of the method as we want to keep backwards compatability.
I wrote you above to make the boolean as flags, and one of the flags should be "update_other_worlds" (or something like that) which will update the other worlds with their own biome.

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
Lemvs: But really, what's the point of it working like that, if the biome could only be changed into one dimension?

API/src/main/java/com/bgsoftware/superiorskyblock/api/island/Island.java
Lemvs: And which flags should exists, set in one dimension and in all dimensions?

Yes, when I tried to do it a few months ago I thought so too, but then I didn't succeed, and now doing it again I did the other way around

src/main/java/com/bgsoftware/superiorskyblock/core/menu/impl/MenuBiomes.java
OmerBenGera, line 1: The reason for not doing it paged menu in the first place is because I was thinking that you'd never get to have more than a page of biomes. Do you really think it's necessary to make the menu paged?

Lemvs: In the menu for 1.16.5 I have already filled the entire page, if I were creating a version for the latest version of Minecraft, there would be more. You can also create your own biomes by datapacks, so yes, someone may need to have more than one page.

The paged menu also allows you to dynamically determine request objects, without it I would have to display biomes, which cannot be set in a given dimension anyway, or create separate patterns, but the number of chars is also limited, and if the more of them are used, the bigger mess is in the file.

src/main/java/com/bgsoftware/superiorskyblock/core/menu/impl/MenuBiomes.java
```diff
@@ -151,4 +305,84 @@ private static boolean convertOldGUI(SuperiorSkyblockPlugin plugin, YamlConfigur
...
+
+                    if (i < chars.length - 1)
+                        newLine.append(" ");
+                }
+
+                newPattern.add(newLine.toString());
+            }
+
+            cfg.set("pattern", newPattern);
+            cfg.set("slots", newChar);
```
OmerBenGera: Will fix the issue with converting:
```suggestion
            cfg.set("slots", String.valueOf(newChar));
```

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
OmerBenGera: I think in the first place I did it so the biomes of other worlds will always be synced.
You are correct that maybe it is not necessary, but someone uses `Island#setBiome(Biome)` expects the same behavior. By adding flags, you can remove this behavior in the plugin's calls - but the API should call the new setBiome using that flag

API/src/main/java/com/bgsoftware/superiorskyblock/api/hooks/MenusProvider.java
Lemvs: Look at this in my new commit, because I don't know if that's what I meant, if not, I need a more detailed explanation 🫤

Lemvs:
Most of it has been fixed, only the UnsupportedOperationException is rather not done as you wanted, but I don't know what exactly is going on, because there is no such situation in this file, so I don't understand it very well.

API/src/main/java/com/bgsoftware/superiorskyblock/api/events/IslandBiomeChangeEvent.java
OmerBenGera: Still missing @deprecated comment

API/src/main/java/com/bgsoftware/superiorskyblock/api/island/Island.java
OmerBenGera: Don't resolve the conversations - they are not fixed.
You should have an UPDATE_WORLD and UPDATE_ALL_WORLDS flags.

API/src/main/java/com/bgsoftware/superiorskyblock/api/enums/DimensionSelectionMode.java
OmerBenGera, line 1: Shouldn't be in API

API/src/main/java/com/bgsoftware/superiorskyblock/api/hooks/MenusProvider.java
```diff
@@ -45,8 +46,21 @@ public interface MenusProvider {
      * @param targetPlayer The player to open the menu for.
      * @param previousMenu The previous menu that was opened, if exists.
      * @param targetIsland The island to change biomes for.
+     * @deprecated See {@link #openBiomes(SuperiorPlayer, ISuperiorMenu, Island, Dimension)}
      */
-    void openBiomes(SuperiorPlayer targetPlayer, @Nullable ISuperiorMenu previousMenu, Island targetIsland);
+    @Deprecated
+    void openBiomes(SuperiorPlayer targetPlayer, @Nullable ISuperiorMenu previousMenu, Island targetIsland) throws UnsupportedOperationException;
```
OmerBenGera:
```suggestion
    default void openBiomes(SuperiorPlayer targetPlayer, @Nullable ISuperiorMenu previousMenu, Island targetIsland) {
    throw new UnsupportedOperationException("openBiomes is not supported")
    }
```

API/src/main/java/com/bgsoftware/superiorskyblock/api/island/DelegateIsland.java
```diff
@@ -858,20 +858,37 @@ public void setPaypal(String paypal) {
     }
 
     @Override
+    @Deprecated
     public Biome getBiome() {
         return this.handle.getBiome();
     }
 
+    public Biome getBiome(Dimension dimension) {
```
OmerBenGera, line 867: Add Override annotation

API/src/main/java/com/bgsoftware/superiorskyblock/api/island/IslandBiomeFlags.java
```diff
@@ -0,0 +1,18 @@
+package com.bgsoftware.superiorskyblock.api.island;
+
+/**
+ * The integer value element annotated with {@link IslandBiomeFlags} represents flags related to what to update
+ * when the biome is changed. It is mainly used within the {@link Island} interface and its methods.
+ */
+public @interface IslandBiomeFlags {
```
OmerBenGera, line 10:
```suggestion
/**
 * The integer value element annotated with {@link IslandBiomeFlags} represents flags related to what to update
 * when the biome is changed. It is mainly used within the {@link Island} interface and its methods.
 */
 @IntType({IslandBiomeFlags.UPDATE_BLOCKS, IslandBiomeFlags.UPDATE_ALL_DIMENSIONS})
public @interface IslandBiomeFlags {
```

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
```diff
@@ -246,7 +247,7 @@ protected IslandCache create() {
     private final AtomicReference<BigDecimal> bonusLevel = new AtomicReference<>(BigDecimal.ZERO);
     private final Map<MissionReference, Counter> completedMissions = new ConcurrentHashMap<>();
     private final Synchronized<IslandChest[]> islandChests = Synchronized.of(createDefaultIslandChests());
-    private final Synchronized<CompletableFuture<Biome>> biomeGetterTask = Synchronized.of(null);
+    private final Synchronized<EnumerateMap<Dimension, CompletableFuture<Biome>>> biomeGetterTasks = Synchronized.of(new EnumerateMap<>(Dimension.values()));
```
OmerBenGera: Create an inner class "IslandBiome" and combine `biomeGetterTasks` and `biomes` into a single Synchronized EnumerateMap.

API/src/main/java/com/bgsoftware/superiorskyblock/api/hooks/MenusProvider.java
```diff
@@ -49,7 +49,9 @@ public interface MenusProvider {
      * @deprecated See {@link #openBiomes(SuperiorPlayer, ISuperiorMenu, Island, Dimension)}
      */
     @Deprecated
-    void openBiomes(SuperiorPlayer targetPlayer, @Nullable ISuperiorMenu previousMenu, Island targetIsland) throws UnsupportedOperationException;
+    default void openBiomes(SuperiorPlayer targetPlayer, @Nullable ISuperiorMenu previousMenu, Island targetIsland) throws UnsupportedOperationException {
```
OmerBenGera: Remove the "throws UnsupportedOperationException" annotation

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
```diff
@@ -1984,20 +1984,24 @@ public Biome getBiome() {
...
-        if (biome == null) {
-            biomeGetterTasks.write(tasksMap -> {
-                tasksMap.computeIfAbsent(dimension, dim -> getBiomeAsyncTask(dimension));
-            });
+        IslandBiome islandBiome = islandBiomes.readAndGet(map -> map.get(dimension));
 
-            return IslandUtils.getDefaultWorldBiome(dimension);
+        if (islandBiome != null && islandBiome.getBiome() != null) {
+            return islandBiome.getBiome();
         }
```
OmerBenGera: Should be inside the readAndGet consumer so it will be inside the lock as well

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
```diff
@@ -2093,7 +2098,7 @@ public void setBiome(Dimension dimension, Biome biome, int flags) {
             for (World registeredWorld : plugin.getGrid().getRegisteredWorlds()) {
                 WorldInfo worldInfo = WorldInfo.of(registeredWorld);
 
-                if (updateAllDimensions || worldInfo.getDimension() == dimension) {
+                if (updateAllDimensions || (updateDimension && worldInfo.getDimension() == dimension)) {
```
OmerBenGera, line 2112: updateDimension is always true in this case. What's the point in "updateDimension" in general?

OmerBenGera (changes requested):

API/src/main/java/com/bgsoftware/superiorskyblock/api/hooks/MenusProvider.java
```diff
@@ -45,8 +46,23 @@ public interface MenusProvider {
      * @param targetPlayer The player to open the menu for.
      * @param previousMenu The previous menu that was opened, if exists.
      * @param targetIsland The island to change biomes for.
+     * @deprecated See {@link #openBiomes(SuperiorPlayer, ISuperiorMenu, Island, Dimension)}
      */
-    void openBiomes(SuperiorPlayer targetPlayer, @Nullable ISuperiorMenu previousMenu, Island targetIsland);
+    @Deprecated
+    default void openBiomes(SuperiorPlayer targetPlayer, @Nullable ISuperiorMenu previousMenu, Island targetIsland) {
+        throw new UnsupportedOperationException("openBiomes is not supported");
```
OmerBenGera: Call openBiomes with default dimension

src/main/java/com/bgsoftware/superiorskyblock/core/menu/impl/MenuBiomes.java
```diff
@@ -151,4 +313,84 @@ private static boolean convertOldGUI(SuperiorSkyblockPlugin plugin, YamlConfigur
...
+
+            cfg.set("pattern", newPattern);
+            cfg.set("slots", String.valueOf(newChar));
+            cfg.set("dimension-selection-mode", "DEFAULT");
+        }
+
+        return true;
+    }
+
+    private static Dimension getDimensionSafe(String dimensionName) {
```
OmerBenGera, line 388: Annotate with `@Nullable`

OmerBenGera (changes requested):

OmerBenGera (changes requested):
Fix conflicts

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
```diff
@@ -2022,46 +2040,88 @@ private CompletableFuture<Biome> getBiomeAsyncTask(CompletableFuture<Biome> curr
...
     @Override
+    @Deprecated
     public void setBiome(Biome biome) {
         setBiome(biome, true);
     }
 
     @Override
+    @Deprecated
     public void setBiome(Biome biome, boolean updateBlocks) {
+        setBiome(plugin.getSettings().getWorlds().getDefaultWorldDimension(), biome,
```
OmerBenGera, line 2059: Should only set `IslandBiomeFlags.UPDATE_BLOCKS` if `updateBlocks` is true

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
```diff
@@ -2049,7 +2049,7 @@ public void setBiome(Biome biome) {
     @Deprecated
     public void setBiome(Biome biome, boolean updateBlocks) {
         setBiome(plugin.getSettings().getWorlds().getDefaultWorldDimension(), biome,
-                IslandBiomeFlags.UPDATE_BLOCKS | IslandBiomeFlags.UPDATE_ALL_DIMENSIONS);
+                updateBlocks ? IslandBiomeFlags.UPDATE_BLOCKS : 0);
```
OmerBenGera: Should always forward `IslandBiomeFlags.UPDATE_ALL_DIMENSIONS`

OmerBenGera (changes requested):

OmerBenGera (approved):

Issue #2572, opened 2025-07-21
New color codes


Lemvs:
### Minecraft's Version

1.21.7

### Plugin's Version

2025.1-b548

### Describe the bug

HEX (&{HEX:xxxxxx}) and MiniMessage format support only works for raw messages (hex format uses approximate color, MiniMessage format is not replaced at all)

Hex usage in complex message:
<img width="511" height="209" alt="Image" src="https://github.com/user-attachments/assets/85e860cb-ecb6-44fe-aff7-5f09c25bee12" />
Hex usage in raw message:
<img width="266" height="41" alt="Image" src="https://github.com/user-attachments/assets/0465f67c-e7d1-4235-aa7a-b9bdeaaf808c" />

### To Reproduce

\-

OmerBenGera:
Fixed

PR #2925, opened 2026-05-02
Fixed MiniMessage support

### LINKED ISSUE #2909: MiniMessage broke message
### Minecraft's Version

paper 1.21.11

### Plugin's Version

dev build #837

### Describe the bug

If I use MiniMessage in language en-US.yml it breaks message to not show on chat.

Image
It means for block_level, block_level_worthless, block_value, block_value_worthless

### To Reproduce

Normal default message work perfect for /is value command but if I use here my MiniMessage it broke and it will not show here anymore. For test I used random text and it works.

My msg to test: <gradient:#87CEEB:#3498DB><bold>WYSPA</bold></gradient> <dark_gray><bold>➜</bold> <gray>Blok <#87CEEB>{0} <gray>nadaje <#87CEEB>{1} <gray>punktów.

Few message above I used it and it works so something bad is here probably.

### Additional Information

Image

between message we can that error with MiniMessage 
Lines:
"level"
"blank message" <- here is problem
"worth brak"
"level brak"
_No response_

Lemvs:
This Pull Request fixes #2909 

Support for MiniMessages has been improved to work correctly for more messages. To achieve this:
- The MiniMessageComponent constructor has been changed: It now accepts the raw message as a String and uses MessageContent.parse(). This replaces the broken findTextComponentContent() method, which iterates over an already parsed Kyori object. Because Kyori splits text into nodes, the old method returned only the first text fragment, completely truncating the rest of the message and losing the original tags.
- The Translator class has been removed: It caused drastic performance degradations. The code used the replaceText() method (which automatically scans all child nodes) and then looped through the children again, creating a "recursion bomb." For large messages, as in /is admin show, this led to server crashes.
- Fixed gradient arguments: The old Translator replaced arguments {...} in the already parsed component tree. This prevented the use of \<rainbow\> or \<gradient\> tags, because Kyori splits such text into individual letters (the {, 0, and } characters were in separate nodes and the replacement didn't see them). Now, using the fixed MessageContent, arguments are replaced in the raw text before parsing, and only then does the finished message go to MINI_MESSAGE.deserialize().
- Removed extra '\n' characters in the code: Manual truncation of trailing newlines has been added to the CmdAdminShow, CmdShow, and CmdValue commands. Unlike the old legacy strings system, MiniMessage doesn't automatically truncate such enters, which resulted in unnecessary, empty lines being sent in chat.
- Improved default language files: Replaced the multi-line YAML string character from | to |-. The missing minus sign caused the YAML parser to automatically append a hidden \n character to the very end of the message. Combined with the new MiniMessage format, this generated a lot of unwanted blank lines in long messages, as in /is admin show.


Lemvs:
I also removed the override for the send() method in the GOT_INVITE message, as it forced the use of the ComplexMessageComponent and prevented the use of the MiniMessageComponent. Instead of two messages, GOT_INVITE and GOT_INVITE_TOOLTIP, there is now a single message in the Interactable Message format ([Wiki](https://wiki.bg-software.com/superiorskyblock/overview/messages#interactable-messages)), with a conversion from the old format to the new one. So now someone who doesn't want this message to be clickable, or wants to create it using MiniMessages tags, can do it (as also suggested in #2909).

Lemvs:
I've also added support for MiniMessage in all components (#2572). To do this, I've added new methods to CustomComponentParser and created equivalents of existing components in MiniMessageHook. I'm not sure if this is done correctly (probably not) - it works, but it could probably be done better. If you have any suggestions, let me know. I was thinking about adding something like ComponentParserProvider. The plugin would have a default provider with the old components, and MiniMessage would create a new one with its components, but I'm not sure if that makes sense. Now it works like this: if MiniMessage doesn't deserialize the component, it falls back to the default components, but you could then just parse the legacy/raw text, MiniMessage has such options.

I also added the ability to choose the BossBar style.

And I removed storing net.kyori.Component in MiniMessages components, because it didn't make sense, even after my previous changes to use it only when args.length == 0, because then placeholders from PlaceholderAPI were not parsed at all (the current code also has this problem, you only replace arguments, which is done very inefficiently anyway).

Tell me what you think @OmerBenGera. When I reported these problems a few months ago, you said you had no idea what to do about it, but maybe seeing how I did it, you'll come up with something, or you'll decide this is enough 😊

OmerBenGera:
I might be doing something wrong, but when I have this:
```
BLOCK_VALUE: '<bold>TEST</bold>{0} <gray>{1}'
```
It doesn't seem to work.

Lemvs:
> I might be doing something wrong, but when I have this:
> 
> ```
> BLOCK_VALUE: '<bold>TEST</bold>{0} <gray>{1}'
> ```
> 
> It doesn't seem to work.

This message isn't sent separately, but combined with BLOCK_LEVEL and sent together, so without knowing how it was set, it's impossible to determine the cause. I replaced all legacy color codes with MiniMessage tags in the lang file, and everything worked fine.

OmerBenGera (approved):

PR #2927, opened 2026-05-07
Added block categories and improved entity categories

Lemvs:
# Changelog # 
- Added the ability to create block categories with permissions for placing, breaking, and interacting with blocks. The old `interactables.yml` file has been replaced by a new `block-categories.yml`, with convert of course. Just as entity categories have built-in categories, block categories include a built-in `ALL` category that contains all blocks for the specific server version.
- Added `skipped-blocks` (and `skipped-entities`) options for built-in block (and entity) categories, allowing users to remove default content from them. This is designed for users who do not wish to require multiple privileges or flags for a specific action.
- Changed the file format for block categories (and entity categories), instead of storing category sections directly, I created `builtin-categories` and `custom-categories` sections. This allows the configuration to be synchronized with the original using `CommentedConfiguration` while making it possible to skip custom categories.
- Removed the `BUILD`, `BREAK`, `SPAWNER_BREAK`, and `VALUABLE_BREAK` privileges from `IslandPrivileges`, as these privileges can now be custom-named within the block categories file.
- Removed the detection of clicking a hoe on a block or anything on `ROOTED_DIRT` as interactions. Only valid actions, such as clicking a hoe on any type of dirt or grass block, are now detected, which is considered placing a block (to align with Bukkit's default recognition of using a shovel on such blocks).
- Removed the requirement to have permission to build when jumping on `FARMLAND` or `TURTLE_EGG`. Instead, the appropriate interaction privilege is required, as before. 
- Removed the requirement for the `TAMED_ANIMAL_DAMAGE` privilege to hit tameable animals. This privilege was intended to prevent killing tamed animals, not tameable animals, which is a significant difference.
- Removed the `ProtectionListener#handleSignColorChange()` method because it was pointless. Changing the text color on a sign isn't placing a block, but a simple interaction, so there's no need to detect it separately. `ProtectionListener#handleBlockInteract()` will detect it as an interaction with the sign, which by default requires the `SIGN_INTERACT` privilege.
- Changed the recognition of a bucket filling situation from placing a block to destroying a block. If we collect something in the bucket, it disappears, so we destroy it, not place it.
- Moved the `ProtectionListener#onBucketDispense()` method to WorldDestructionListener, as that's where the listener that detects whether we're damaging the terrain outside the island should be.
- Standardized the code in `RegionManagerServiceImpl` and `ProtectionListener` by using more {}, wrapping long lines, and adding blank lines when there was too much code next to each other.

Unifying the category system by also introducing categories for blocks will allow for easier implementation of group limits in the future. That's why I suggested this to you earlier in one of the issues, and since you weren't willing to take it on, I decided to try it myself. Let me know how it went and what improvements I could make @OmerBenGera 🥹


nivcoo:
I didn't see this PR, here https://github.com/BG-Software-LLC/SuperiorSkyblock2/pull/3035 i saw that the tamed permission doesn't target tamed animals but tameable so i fixed it. But here you say "Removed the requirement for the TAMED_ANIMAL_DAMAGE privilege to hit tameable animals. This privilege was intended to prevent killing tamed animals, not tameable animals, which is a significant difference."

Do you fix it too ?

src/main/java/com/bgsoftware/superiorskyblock/world/BukkitEntities.java
```diff
@@ -164,10 +164,12 @@ public static boolean isNautilus(EntityType entityType) {
 
     public static List<EntityCategory> getCategories(Entity entity) {
```
Lemvs: @nivcoo yes.

Lemvs:
> I didn't see this PR, here #3035 i saw that the tamed permission doesn't target tamed animals but tameable so i fixed it. But here you say "Removed the requirement for the TAMED_ANIMAL_DAMAGE privilege to hit tameable animals. This privilege was intended to prevent killing tamed animals, not tameable animals, which is a significant difference."
> 
> Do you fix it too ?


I’ve tagged you in the code. I didn’t change the category name because it does contain tameable animals after all; I simply remove the category if the mob isn’t tamed. It’s impossible to create a list of "tamed animals types" because that is an attribute of a specific mob, so I would keep the current name regardless.

nivcoo:
I do it, i create the TAMED category and only add it if it's tamed, because yes we can still use tameable but idk if remove it for tamed mobs is the best way to do it

Lemvs:
> I do it, i create the TAMED category and only add it if it's tamed, because yes we can still use tameable but idk if remove it for tamed mobs is the best way to do it

It amounts to the same thing, you’re just doing it in reverse order - I’m removing categories while you’re adding them, though that did require you to introduce different constructors. My pull requests likely won’t be accepted anytime soon 🫠, so they’ll probably go with your solution, I can always change it later if necessary 🤔

nivcoo:
I guess the best way is merge our system to filtrate if user wants, filter TAMED animal or globally TAMEABLE for exemple filter only animal not tamed but can be tamed so add tameable tag when we can and add TAMED tag when the animal is tamed

nivcoo:
Is this PR add the feature to add block limit to group of materials ? Like all "sign" etc ?

Lemvs:
> Is this PR add the feature to add block limit to group of materials ? Like all "sign" etc ? 

No, first I need to have this pull request approved to do the same as there https://github.com/BG-Software-LLC/SuperiorSkyblock2/pull/3020 

I wrote about it in the description of that pull request

nivcoo:
Ok that's good

Lemvs:
> Ok that's good

There won't be any problems with it, because I'll have to do exactly the same as there only for blocks, but I have to have a block category system first to continue, and I don't want to make more changes in this pull request, because a lot has already changed here

Issue #2888, opened 2026-03-12
top Placeholders cannot be used when a custom SortingType is registered.


Wakkaer2:
### Minecraft's Version

This server is running Paper version 1.21.11-69-main@94d0c97 (2025-12-30T20:33:30Z) (Implementing API version 1.21.11-R0.1-SNAPSHOT)

### Plugin's Version

SuperiorSkyblock2 version 2025.2.1-b820

### Describe the bug

```
[18:15:42 ERROR]: Command exception: /papi parse --null %superior_island_top_completed_1%
org.bukkit.command.CommandException: Unhandled exception executing command 'papi' in plugin PlaceholderAPI v2.12.2
        at org.bukkit.command.PluginCommand.execute(PluginCommand.java:47) ~[paper-api-1.21.11-R0.1-SNAPSHOT.jar:?]
        at io.papermc.paper.command.brigadier.bukkit.BukkitCommandNode$BukkitBrigCommand.run(BukkitCommandNode.java:83) ~[paper-1.21.11.jar:1.21.11-69-94d0c97]
        at com.mojang.brigadier.context.ContextChain.runExecutable(ContextChain.java:73) ~[brigadier-1.3.10.jar:?]
        at net.minecraft.commands.execution.tasks.ExecuteCommand.execute(ExecuteCommand.java:30) ~[paper-1.21.11.jar:1.21.11-69-94d0c97]
        at net.minecraft.commands.execution.tasks.ExecuteCommand.execute(ExecuteCommand.java:13) ~[paper-1.21.11.jar:1.21.11-69-94d0c97]
        at net.minecraft.commands.execution.UnboundEntryAction.lambda$bind$0(UnboundEntryAction.java:8) ~[paper-1.21.11.jar:1.21.11-69-94d0c97]
        at net.minecraft.commands.execution.CommandQueueEntry.execute(CommandQueueEntry.java:5) ~[paper-1.21.11.jar:1.21.11-69-94d0c97]
        at net.minecraft.commands.execution.ExecutionContext.runCommandQueue(ExecutionContext.java:104) ~[paper-1.21.11.jar:1.21.11-69-94d0c97]
        at net.minecraft.commands.Commands.executeCommandInContext(Commands.java:469) ~[paper-1.21.11.jar:1.21.11-69-94d0c97]
[...]

PR #2935, opened 2026-05-18
Fix top placeholder regex regression for value/leader/raw subkeys

djoey123:
Commit efbdfcea6 (PR #2888) replaced the per-SortingType prefix patterns with a single greedy regex "(.+)_(.+)". This split placeholders like "level_value_1" into ("level_value", "1") instead of ("level", "value_1"), making SortingType.getByName return null and the placeholder resolve to "-" for every value/value_raw/value_format/leader variant.

Replace the regex with a longest-prefix match against the registered SortingType names. Custom sorting types whose name contains an underscore also resolve correctly because longer names are tried first.

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/service/placeholders/PlaceholdersServiceImpl.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/service/placeholders/PlaceholdersServiceImpl.java
@@ -35,6 +35,8 @@ import org.bukkit.potion.PotionEffectType;
 
 import java.math.BigDecimal;
 import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.LinkedList;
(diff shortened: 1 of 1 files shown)


src/main/java/com/bgsoftware/superiorskyblock/service/placeholders/PlaceholdersServiceImpl.java
```diff
@@ -823,20 +824,35 @@ private static Optional<String> handlePermissionRolesPlaceholder(@NotNull Island
...
+        String placeholderValue = null;
+
+        for (SortingType candidate : orderedTypes) {
+            String candidateName = candidate.getName().toLowerCase(Locale.ENGLISH);
+            if (subPlaceholderLower.startsWith(candidateName + "_")) {
+                sortingType = candidate;
+                placeholderValue = subPlaceholder.substring(candidateName.length() + 1);
+                break;
+            }
+        }
```
OmerBenGera: A better fix would be to just use lazy regex? `(.+?)_(.+)`
This fix will be heavy performance

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/service/placeholders/PlaceholdersServiceImpl.java
djoey123: I have changed it, but it might not work as you expected before the change you made to support custom placeholders from others, as I do not know what you tested this on.

What I could also have done was use caching instead of walking over it again every time, but I think you know better how much impact that will have.

OmerBenGera (approved):

PR #2949, opened 2026-06-10
Fix /is bal showing command sender name on visited islands

GenOpenDiesel:
## Summary
- Fix `getIslandWhereStanding` to return the island owner as the target player when the command sender is standing on another island.
- Harden `/is bal` message selection so the balance output uses the island owner or island name instead of the visitor when the resolved target player does not match the island owner.

## Problem
When a player teleported to another island and ran `/is bal` without arguments, the balance amount was correct but the message still showed the visitor's name instead of the visited island owner.

## Test plan
- [ ] Run `/is bal` on your own island and confirm the "your island" message is shown with the correct balance.
- [ ] Teleport to another player's island warp and run `/is bal`; confirm the message shows the visited island owner/name with that island's balance.
- [ ] Run `/is bal <player>` and `/is bal <island-name>` and confirm existing targeted balance output still works.

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/commands/arguments/CommandArguments.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/commands/arguments/CommandArguments.java
@@ -131,7 +131,9 @@ public class CommandArguments {
         if (island == null)
             Message.INVALID_ISLAND.send(sender);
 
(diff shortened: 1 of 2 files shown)


OmerBenGera:
I pushed a fix in latest dev build, thanks for the PR!

Issue #1625, opened 2023-03-05
RoseStacker Spawners cannot be disabled with is settings


George-Rowland:
### Minecraft's Version

This server is running Paper version git-Paper-307 (MC: 1.19.2) (Implementing API version 1.19.2-R0.1-SNAPSHOT) (Git: 476ef25)
You are running the latest version

### Plugin's Version

[12:07:30 INFO]: SuperiorSkyblock2 version 2022.9
[12:07:30 INFO]: Feature packed Skyblock core.
[12:07:30 INFO]: Website: https://bg-software.com/
[12:07:30 INFO]: Author: Ome_R

### Describe the bug

 RoseStacker Spanwers are still spawning mobs even when toggled in /is settings.
See for reproduction and video: https://youtu.be/WjZm0F4HUqs

Spawners are set to use RoseStacker: https://prnt.sc/bSMDkf4NrfQh

### To Reproduce

Place a RoseStacker spawner, toggle mob spawning off in /is settings, mobs still spawn

### Additional Information

https://youtu.be/WjZm0F4HUqs

OmerBenGera:
Contact RoseStacker regarding this.
SuperiorSkyblock disables the SpawnerSpawnEvent, and I suspect they do not call it properly.

George-Rowland:
I have passed this on to RoseStacker support and will keep you updated if applicable

OmerBenGera:
> I have passed this on to RoseStacker support and will keep you updated if applicable

Awesome!

PR #2950, opened 2026-06-13
Enforce spawner-spawning island flag for RoseStacker

djoey123:
RoseStacker spawns mobs through a custom NMS path that, by default (bypass-region-spawning-rules: true), does not fire a Bukkit CreatureSpawnEvent, and also merges spawns directly into nearby entity stacks without spawning a new entity. Both paths bypass SuperiorSkyblock's CreatureSpawnEvent-based spawner-spawning flag check, so disabling spawner mob spawning via /is settings had no effect (issue #1625).

Listen to RoseStacker's cancellable PreStackedSpawnerSpawnEvent, which gates every spawn path, and cancel it when the island has the spawner-spawning flag disabled for the spawned entity category.

This event (and StackedSpawnerTile#getSpawnedType) only exist in newer RoseStacker versions, which also ship as Java 17 classes. To keep supporting older Java 8 RoseStacker versions without failing to load, the event is accessed entirely through reflection and the listener is only registered when the event class is present at runtime. getStack/setCancelled are bound to the StackEvent superclass because ReflectMethod only resolves declared methods. On older versions the regular CreatureSpawnEvent listener keeps handling the flag as before.


djoey123:
Tested and is working :)

OmerBenGera (changes requested):
I am thinking about creating a new module for the newer versions of RoseStacker, so we don't need to do all the implementation through reflection. Then, you can choose which RoseStacker modules to use depends on the version of the plugin

djoey123:
Also tested and working :)

PR #2962, opened 2026-07-08
Added Folia Support.

KeyHeld:
Added Folia Support.


OmerBenGera:
Did you test it?
Does Folia support world api?

OmerBenGera:
I tested it and it doesn't work.

PR #2963, opened 2026-07-08
improv. add registration listeners for island flags and privileges

nivcoo:
Hello, i see that with wildstacekr i had the issue so i fix it directly into SSB2. Actually the cache is loaded on plugin start so if flag or permission was registered after that the cache doesn't count them. So i used listerner to allow that. I tried it and it fix correctly the issue.


OmerBenGera:
Which cache?
Also, use bukkit events for that

nivcoo:
The "DEFAULT_FLAGS_CACHE"

nivcoo:
I migrate to bukkit event in API

OmerBenGera:
My bad - I didn't see you register listeners for internal use only, and bukkit events are not the way to handle it.
The best solution for this is to make a new class called "EnumerateSet" which extends EnumerateSet. It will hold a set of island flag names that weren't parsed, and will test against island flag names as well. When it matches a string in this set, it will add the flag for cached results in future lookups. Then, replace all the instances where EnumerateSet stores cached values.

nivcoo:
I will migrate to it but i'm pretty sure that my first attempt was a good structure but i will use it yes

API/src/main/java/com/bgsoftware/superiorskyblock/api/island/IslandFlag.java
```diff
@@ -60,7 +60,8 @@ public static void register(String name) {
 
         Preconditions.checkState(!islandFlags.containsKey(name), "IslandFlag with the name " + name + " already exists.");
 
-        islandFlags.put(name, new IslandFlag(name));
+        IslandFlag islandFlag = new IslandFlag(name);
+        islandFlags.put(name, islandFlag);
```
OmerBenGera: Revert this change

API/src/main/java/com/bgsoftware/superiorskyblock/api/island/IslandPrivilege.java
```diff
@@ -92,7 +92,8 @@ public static void register(String name, Type type) {
 
         Preconditions.checkState(!islandPrivileges.containsKey(name), "IslandPrivilege with the name " + name + " already exists.");
 
-        islandPrivileges.put(name, new IslandPrivilege(name, type));
+        IslandPrivilege islandPrivilege = new IslandPrivilege(name, type);
+        islandPrivileges.put(name, islandPrivilege);
```
OmerBenGera: Revert this change

src/main/java/com/bgsoftware/superiorskyblock/core/collections/UnparsedEnumerateSet.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    public boolean contains(K key) {
+        if (super.contains(key))
+            return true;
+
+        if (unparsedNames.remove(normalizeName(nameProvider.apply(key)))) {
+            add(key);
+            return true;
+        }
```
OmerBenGera: Optimization: if unparsedNames is empty, don't even call "nameProvider" and just fast-fail

src/main/java/com/bgsoftware/superiorskyblock/core/collections/UnparsedEnumerateSet.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Set;
+import java.util.function.Function;
+
+public class UnparsedEnumerateSet<K extends Enumerable> extends EnumerateSet<K> {
+
+    private final Set<String> unparsedNames = new HashSet<>();
+    private final Function<String, K> parser;
+    private final Function<K, String> nameProvider;
+
+    public UnparsedEnumerateSet(Collection<K> enumerables, Function<String, K> parser, Function<K, String> nameProvider) {
```
OmerBenGera: Instead of providing two functions, make `UnparsedEnumerateSet` an abstract class and let the callers implement these functions.

OmerBenGera (changes requested):

nivcoo:
Is it good ?

OmerBenGera:
Perfect.
Does it work with WildStacker now? I just want you to make sure the changes are still fixing the bug

OmerBenGera (approved):

nivcoo:
Yes it works i try it

OmerBenGera (approved):

src/main/java/com/bgsoftware/superiorskyblock/core/collections/UnparsedEnumerateSet.java
```diff
@@ -0,0 +1,53 @@
...
+    public UnparsedEnumerateSet(Collection<K> enumerables) {
+        super(enumerables);
+    }
+
+    public boolean addName(String name) {
+        try {
+            return add(parseName(name));
+        } catch (Throwable ignored) {
+            if (name == null)
+                return false;
```
OmerBenGera: Can be checked at the top of the method

OmerBenGera (changes requested):

nivcoo:
good

OmerBenGera (approved):

PR #2966, opened 2026-07-10
feat(generator): support custom datapacks and disable void generator

babycrashdev:
This Pull Request introduces compatibility with custom world generators (datapacks/vanilla generation) and pre-loaded worlds.

1. **Option to Disable Void Generator**:
   * Added `use-void-generator` configuration option per-dimension in `config.yml` (defaults to `true`).
   * Updated `WorldsProvider_Default.java` to respect this option, allowing worlds to use vanilla or datapack generators instead of forcing the void generator.
   * Modified `GridManagerImpl.java` to skip the post-creation chunk clearing task if `use-void-generator` is set to `false`, preventing naturally generated terrain around islands from being voided.
   * Exposed `isUseVoidGenerator()` in the `DimensionConfig` API.

2. **Pre-loaded Worlds Compatibility**:
   * Removed the hard `RuntimeException` in `WorldsProvider_Default.java` when a world is already loaded.
   * SuperiorSkyblock will now print a warning in the console and gracefully hook into the existing world instead of crashing the server.

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/config/section/WorldsSection.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/config/section/WorldsSection.java
@@ -13,6 +13,7 @@ import com.bgsoftware.superiorskyblock.core.serialization.Serializers;
 import org.bukkit.PortalType;
 import org.bukkit.World;
 import org.bukkit.configuration.ConfigurationSection;
+import com.bgsoftware.superiorskyblock.api.enums.GeneratorHint;
 
 import java.util.Collections;
 import java.util.EnumMap;
@@ -74,7 +75,7 @@ public class WorldsSection extends SettingsContainerHolder implements SettingsMa
     public static abstract class BaseDimensionConfig implements DimensionConfig {
 
         private final boolean isEnabled;
-        private final boolean isUseVoidGenerator;
+        private final GeneratorHint generatorHint;
         private final boolean isUnlocked;
         private final boolean isSchematicOffset;
         private final String biome;
@@ -83,7 +84,15 @@ public class WorldsSection extends SettingsContainerHolder implements SettingsMa
 
         protected BaseDimensionConfig(ConfigurationSection section, Dimension dimension, String defaultName) {
             this.isEnabled = section.getBoolean("enabled");
-            this.isUseVoidGenerator = section.getBoolean("use-void-generator", true);
+            String generatorHintStr = section.getString("generator-hint", "VOID");
+            GeneratorHint generatorHint;
+            try {
+                generatorHint = GeneratorHint.valueOf(generatorHintStr.toUpperCase());
+            } catch (Exception error) {
+                Log.warnFromFile("config.yml", "Invalid generator hint ", generatorHintStr, " - using VOID instead.");
+                generatorHint = GeneratorHint.VOID;
+            }
+            this.generatorHint = generatorHint;
             this.isUnlocked = section.getBoolean("unlock");
             this.isSchematicOffset = section.getBoolean("schematic-offset");
             this.biome = section.getString("biome");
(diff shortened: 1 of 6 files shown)


OmerBenGera:
One of the most best things with SSB that if it knows the generator is a void-type generator, the reseting of chunks etc is done really really fast, and it skips entirely the generation of chunks.
Therefore, I don't like the "isVoidGenerator" thing. Instead, change this to something like "generator-hint", which will be a hint for the plugin for the type of generator used for the world - this will be an enum of different types of generators (VOID, CUSTOM). Then, instead of checking in the plugin if the generator is "IslandGeneratorImpl", it will check if the hint is of type VOID.

babycrashdev:
Hello,

I'm not entirely sure I fully understood your request, but here's what has been implemented:
* **Generator Hint enum:** Introduced a new `GeneratorHint` enum (`VOID`, `CUSTOM`) in the API.
* **API update:** Replaced `isUseVoidGenerator()` with `getGeneratorHint()` in `DimensionConfig`.

### Logic updates
The relevant checks have been updated accordingly:
* The fast chunk reset logic in `GridManagerImpl`.
* The world loading logic in `WorldsProvider_Default`.

Both now rely on `getGeneratorHint() == GeneratorHint.VOID` to determine whether the optimized void-generation path should be used.

### Default configuration
The default `config.yml` template has also been updated to document and use `generator-hint: VOID` by default.

src/main/java/com/bgsoftware/superiorskyblock/config/section/WorldsSection.java
```diff
@@ -83,7 +84,15 @@ public static abstract class BaseDimensionConfig implements DimensionConfig {
...
             this.isEnabled = section.getBoolean("enabled");
-            this.isUseVoidGenerator = section.getBoolean("use-void-generator", true);
+            String generatorHintStr = section.getString("generator-hint", "VOID");
+            GeneratorHint generatorHint;
+            try {
+                generatorHint = GeneratorHint.valueOf(generatorHintStr.toUpperCase());
+            } catch (Exception error) {
+                Log.warnFromFile("config.yml", "Invalid generator hint ", generatorHintStr, " - using VOID instead.");
+                generatorHint = GeneratorHint.VOID;
+            }
```
OmerBenGera, line 94:
```suggestion
            try {
                generatorHint = GeneratorHint.valueOf(generatorHintStr.toUpperCase(Locale.ENGLISH));
            } catch (Exception error) {
                Log.warnFromFile("config.yml", "Invalid generator hint ", generatorHintStr, " - using CUSTOM instead.");
                generatorHint = GeneratorHint.CUSTOM;
            }
```

src/main/java/com/bgsoftware/superiorskyblock/island/GridManagerImpl.java
```diff
@@ -377,7 +379,11 @@ private void createIslandInternalOnSuccessCallback(IslandBuilderImpl builder, Bi
                     if (result) {
                         if (affectedChunks != null) {
                             BukkitExecutor.sync(() -> {
-                                IslandUtils.resetChunksExcludedFromList(island, affectedChunks);
+                                SettingsManager.Worlds.DimensionConfig dimensionConfig =
+                                        plugin.getSettings().getWorlds().getDimensionConfig(defaultDimension);
+                                if (dimensionConfig == null || dimensionConfig.getGeneratorHint() == GeneratorHint.VOID) {
+                                    IslandUtils.resetChunksExcludedFromList(island, affectedChunks);
+                                }
```
OmerBenGera, line 386: You should always reset chunks

OmerBenGera (changes requested):
This is a partial, non full support for custom generators.
Take a look at all the xrefs of "IslandGenerator", you'll see where I use it. For example, in NMSChunksImpl#removeBlocks - instead of checking if the chunkGenerator is not an instanceof IslandGenerator, you should check if the generator hint is non-void.

PR #2967, opened 2026-07-10
Added messages provider

Lemvs:
# Changelog #
- Refactored MessagesServiceImpl.CustomComponentParser into MessagesProvider to simplify the use of new parsers, the entire implementation of the MiniMessageHook (which based on CustomComponentParser) has been converted into MessagesProvider_MiniMessage. If the server supports MiniMessage, this provider is used regardless of whether the message contains legacy color codes or MiniMessage tags, as the deserializer now handles both cases. If the server does not support MiniMessage, MessagesProvider_Default is used, parsing messages as before.
- Added MiniMessage support for /island admin title and /island admin titleall.
- Moved the logic for creating BaseComponents for ComplexMessageComponents into that class itself, rather than handling it within MultipleMessagesComponent.
- Added support for BossBar styles to the default BossBarComponent as well.
- Fixed the creation of the TitleComponent by MiniMessage. Newer versions of Adventure (such as the one used by Paper 26.2) no longer have the Title.Times.of() method, they replaced it to Title.Times.times(), so the correct one must be selected.

I’m not sure if I did everything correctly, but the previous CustomComponentParser seemed very inconvenient to use, and the Provider seemed like a simpler option for using MiniMessage. Let me know if I should change anything.


Hooks/MiniMessage/src/main/java/com/bgsoftware/superiorskyblock/external/messages/MessagesProvider_MiniMessage.java
```diff
@@ -241,27 +196,56 @@ public String getMessage(Object... args) {
...
+            this.messageContent.getContent(player, args).ifPresent(message ->
+                sender.sendMessage(deserialize(message)));
+        }
+
+    }
+
+    // Adventure 5.x renamed the Title.Times factory method from 'of' to 'times',
+    // breaking binary compatibility with Adventure 4.x. Resolve the method
+    // reflectively so a single build works on both versions.
+    private static final ReflectMethod<Title.Times> TITLE_TIMES_FACTORY = getTitleTimesFactory();
```
OmerBenGera: Should be at the top of the class

Hooks/MiniMessage/src/main/java/com/bgsoftware/superiorskyblock/external/messages/MessagesProvider_MiniMessage.java
```diff
@@ -241,27 +196,56 @@ public String getMessage(Object... args) {
...
+        }
+
+        throw new ExceptionInInitializerError("Couldn't find a method to create a Title for MiniMessage.");
     }
 
-    private static class MiniMessageTitleComponent implements IMessageComponent {
+    private static Title.Times createTimes(int fadeIn, int stay, int fadeOut) {
+        return TITLE_TIMES_FACTORY.invoke(null, Duration.ofMillis(fadeIn * 50L),
+                Duration.ofMillis(stay * 50L), Duration.ofMillis(fadeOut * 50L));
+    }
```
OmerBenGera: Move above all classes, below the class's methods

API/src/main/java/com/bgsoftware/superiorskyblock/api/hooks/MessagesProvider.java
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Creates an {@link IMessageComponent} representing a boss bar.
+     *
+     * @param name The boss bar name.
+     * @param color The boss bar color.
+     * @param style The boss bar style.
+     * @param duration The boss bar duration, in ticks.
+     * @return The created message component.
+     */
+    IMessageComponent createBossBarComponent(String name, String color, String style, int duration);
```
OmerBenGera: I think `style` and `color` should be the enums from the API

src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminTitle.java
```diff
@@ -104,19 +92,8 @@ public void execute(SuperiorSkyblockPlugin plugin, CommandSender sender, Superio
...
-            subtitle = placeholdersService.get().parsePlaceholders(player, subtitle);
-
-        plugin.getNMSPlayers().sendTitle(player,
-                title == null ? null : Formatters.COLOR_FORMATTER.format(title),
-                subtitle == null ? null : Formatters.COLOR_FORMATTER.format(subtitle),
-                fadeIn.getNumber(),
-                duration.getNumber(),
-                fadeOut.getNumber());
+        plugin.getProviders().getMessagesProvider().createTitleComponent(title, subtitle,
+                fadeIn.getNumber(), duration.getNumber(), fadeOut.getNumber()).sendMessage(targetPlayer.asPlayer());
```
OmerBenGera: I think the correct way would be to use MessagesService and use its Builder rather than accessing MessagesProvider directly. Unlike other providers, MessagesProvider should probably be used internally in MessagesService.

API/src/main/java/com/bgsoftware/superiorskyblock/api/hooks/MessagesProvider.java
OmerBenGera, line 1: Currently let's make MessagesProvider for internal use only (aka, don't expose it to the API yet)

OmerBenGera (changes requested):

API/src/main/java/com/bgsoftware/superiorskyblock/api/hooks/MessagesProvider.java
Lemvs: Unfortunately, but the bukkit bossbar style has different values than the minimessage overlay, in the style it is 'solid', while in the overlay there is 'progress', so for consistency I set both arguments as a string 😕

And instead of SEGMENTED_..., there are NOTCHED_..., so idk how to do it correctly, maybe use our BossBar.Style enum anyway and just make switch case in MiniMessage createBossBarComponent method to select appropiate overlay from it?

src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminTitle.java
Lemvs: Yeah, I did the same thing in SIsland.sendTitle(), Message.CUSTOM.send() and MultipleComponents.parseSection().

API/src/main/java/com/bgsoftware/superiorskyblock/api/hooks/MessagesProvider.java
Lemvs: Sorry @OmerBenGera, but I didn't notice you had marked this as resolved, but I went ahead and changed the signature anyway, by adding a few helper options to those enums for that purpose. However, I'm not sure if I should be doing that sort of thing in api, so let me know how this should ultimately be handled.

OmerBenGera: You're right, it shouldn't be in the API haha
Let's move this to the MessagesProvider themselves, and they should handle the styles.
I think that's totally reasonable to have "duplicated enums" for handling both Spigot and Paper names.

API/src/main/java/com/bgsoftware/superiorskyblock/api/service/bossbar/BossBar.java
```diff
@@ -36,17 +38,54 @@ enum Color {
...
-        WHITE
+        WHITE;
+
+        public static Color getSafe(String name) {
+            try {
+                return Color.valueOf(name.toUpperCase(Locale.ENGLISH));
+            } catch (IllegalArgumentException exception) {
+                return PINK;
+            }
+        }
```
OmerBenGera: Remove this. In the code, use EnumHelper for that.

API/src/main/java/com/bgsoftware/superiorskyblock/api/service/bossbar/BossBar.java
```diff
@@ -36,17 +38,54 @@ enum Color {
...
+                case "SEGMENTED_12":
+                case "NOTCHED_12":
+                    return SEGMENTED_12;
+                case "SEGMENTED_20":
+                case "NOTCHED_20":
+                    return SEGMENTED_20;
+                default:
+                    return SOLID;
+            }
+        }
```
OmerBenGera: Convert to regular enums with duplicated names (aka, both `SEGMENTED_20` and `NOTCHED_20` will be a valid enums), and each MessagesProvider will map them to the underline spigot/paper enums.

Hooks/MiniMessage/src/main/java/com/bgsoftware/superiorskyblock/external/messages/MessagesProvider_MiniMessage.java
```diff
@@ -1,145 +1,108 @@
...
-import org.bukkit.configuration.file.YamlConfiguration;
 import org.bukkit.entity.Player;
 
 import java.time.Duration;
 import java.util.HashSet;
-import java.util.Optional;
 import java.util.Set;
 
-public class MiniMessageHook {
+@SuppressWarnings("unused")
```
OmerBenGera: Remove `SuppressWarnings`

src/main/java/com/bgsoftware/superiorskyblock/external/messages/MessagesProvider_Default.java
OmerBenGera, line 1: Let's move all the specific-type components (ActionBarComponent, BossBarComponent, etc) to be sub-classes of this provider.
Only components that are used as utilities (for example, "MultipleComponents") should be exposed and be used anywhere.

OmerBenGera (changes requested):

Lemvs:
@OmerBenGera I also have a question, because right now, I’m using the Color Formatter separately in each MessagesProvider. Would it be better to handle this directly in MultipleComponents.parseSection() and MessagesService.parseComponent()? After all, I have to format the colors anyway, even for MiniMessage, so that its original deserializer throws an exception, allowing LegacyComponentSerializer to perform the deserialization.

OmerBenGera:
> MessagesService

Yes, you want to format color codes directly as soon as possible - this is because we don't want to format the color codes everytime the message is sent. Furthermore, I'd avoid the exception when possible - store whether the message has "legacy colors" and if it does - use the legacy formatter.

Lemvs:
> > MessagesService
> 
> Yes, you want to format color codes directly as soon as possible - this is because we don't want to format the color codes everytime the message is sent. Furthermore, I'd avoid the exception when possible - store whether the message has "legacy colors" and if it does - use the legacy formatter.

Hmm, I’m not sure if keeping that information is worth it, because what if someone provides an argument or uses a placeholder that results in color codes after replacement? That information probably wouldn't be of much use, you'd still have to check it every time anyway, for such situations :/

OmerBenGera:
> > > MessagesService
> > 
> > 
> > Yes, you want to format color codes directly as soon as possible - this is because we don't want to format the color codes everytime the message is sent. Furthermore, I'd avoid the exception when possible - store whether the message has "legacy colors" and if it does - use the legacy formatter.
> 
> Hmm, I’m not sure if keeping that information is worth it, because what if someone provides an argument or uses a placeholder that results in color codes after replacement? That information probably wouldn't be of much use, you'd still have to check it every time anyway, for such situations :/

In this case the exception will be thrown, then you'll use the legacy formatter as well.
This is a "fast check" for optimization purposes, not functionality.

Lemvs:
> In this case the exception will be thrown, then you'll use the legacy formatter as well. This is a "fast check" for optimization purposes, not functionality.

Right, done. See if this is what you had in mind. MessageContent is created once, so it is easy to store it there.

src/main/java/com/bgsoftware/superiorskyblock/core/messages/MessageContent.java
```diff
@@ -61,22 +63,43 @@ public static MessageContent parse(String content) {
                 previousPart.append(matcher.group());
             }
 
-            if (previousPart.length() > 0)
-                parts.add(new StaticPart(previousPart.toString()));
-            if (argumentPart != null)
+            if (previousPart.length() > 0) {
+                String previous = previousPart.toString();
+
+                if (previous.indexOf('§') >= 0) {
```
OmerBenGera: Use `ChatColor.COLOR_CHAR`

src/main/java/com/bgsoftware/superiorskyblock/core/messages/MessageContent.java
```diff
@@ -61,22 +63,43 @@ public static MessageContent parse(String content) {
...
 
             lastPartIdx = matcher.end();
         }
 
-        if (lastPartIdx < content.length())
-            parts.add(new StaticPart(content.substring(lastPartIdx)));
+        if (lastPartIdx < content.length()) {
+            String remaining = content.substring(lastPartIdx);
+
+            if (remaining.indexOf('§') >= 0) {
```
OmerBenGera: Use `ChatColor.COLOR_CHAR`

OmerBenGera (changes requested):

OmerBenGera:
Last thing - either the message providers handle colors, or the callers handle it. Don't mix it, or translate colors twice.
Go through all the implementations and make sure they all do it in the same place.

OmerBenGera:
I noticed a potential bug - I think empty messages are sent, even tho in the default implementation (or at least in the latest dev build) they aren't.

Lemvs:
> Last thing - either the message providers handle colors, or the callers handle it. Don't mix it, or translate colors twice. Go through all the implementations and make sure they all do it in the same place.

Currently, Color Formatter is used in MessagesServiceImpl.parseComponent(), which implies its use in MultipleComponents.parseSection() and ComplexComponents.parseBaseComponents() as well, but I no longer use it directly in providers at all.

OmerBenGera (approved):

PR #2976, opened 2026-07-12
Fixed sending messages to island chats when they hadn't been manually enabled before

Lemvs:
Since you told me to remove providing the current ChatState in PlayerChangeChatStateEvent, a problem arose where using `/is localchat <message>` or `/is teamchat <message>` didn't work correctly, if a player hadn't previously enabled that chat mode using the command without the message argument, the message would attempt to send to the wrong chat channel.


src/main/java/com/bgsoftware/superiorskyblock/commands/player/CmdLocalChat.java
OmerBenGera, line 1: Using `/is localchat <message>` or `/is teamchat <message>` shouldn't change the chat state of the player.
The fix should be adding a parameter to `handleIslandChat` that indicates what ChatState to use, rather than always use the player's chat state. Another thing I noticed is that `handleIslandChat` calls `IslandChatEvent` even if the chat state is neither TeamChat or LocalChat

OmerBenGera (changes requested):

Lemvs:
@OmerBenGera I did what you wrote in the requested changes, but analyzing it again now, it doesn't make sense. If you wrote that I should not add ChatState to IslandChatEvent, because you can take ChatState from the player's object, how can someone handle IslandChatEvent, if using /is localchat <message>, the player can have a completely different ChatState set (the same for teamchat), so how can someone using the API determine which chat channel the player wanted to send a message to? Either you have to call PlayerChangeChatStateEvent also when sending a message through the command argument, or you have to give the used ChatState in the IslandChatEvent constructor, without it, this feature does not make sense, because if someone prefers to send a message themselves to for example be able to add rank prefixes, or monitor spam or bad words, he currently cannot do it correctly. When there was one chat, there was no problem, because it was known that the player sent a message to the team, even if his hasTeamChatEnabled() method returned false, but now that there are more chats, such a system will not work correctly

OmerBenGera:
> @OmerBenGera I did what you wrote in the requested changes, but analyzing it again now, it doesn't make sense. If you wrote that I should not add ChatState to IslandChatEvent, because you can take ChatState from the player's object, how can someone handle IslandChatEvent, if using /is localchat , the player can have a completely different ChatState set (the same for teamchat), so how can someone using the API determine which chat channel the player wanted to send a message to? Either you have to call PlayerChangeChatStateEvent also when sending a message through the command argument, or you have to give the used ChatState in the IslandChatEvent constructor, without it, this feature does not make sense, because if someone prefers to send a message themselves to for example be able to add rank prefixes, or monitor spam or bad words, he currently cannot do it correctly. When there was one chat, there was no problem, because it was known that the player sent a message to the team, even if his hasTeamChatEnabled() method returned false, but now that there are more chats, such a system will not work correctly

Add it as a parameter to the event

OmerBenGera (approved):

OmerBenGera (approved):

PR #2982, opened 2026-07-14
Added MiniMessage support in menus

Lemvs:
# Changelog # 
- Added `InventoryProvider`, which handles setting the display name and lore of an item's meta, as well as creating the inventory. The default variant - `InventoryProvider_Default`, performs these tasks just as before, whereas `InventoryProvider_MiniMessage` uses components (similar to `MessagesProvider_MiniMessage`).
- ItemBuilder no longer modifies the display name and lore directly on the item meta object. Instead, it stores them separately, and they are now applied to the item meta once, during the `build()` method, using the inventory provider to allow the use of MiniMessage in it. Additionaly, I cleaned up this class; it's more consistent now.
- `RegularInventoryMenuLayoutImpl`, `SIslandChest` and `StackedBlocksDepositMenu` no longer create menus themselves using `Bukkit.createInventory()`, but instead use a menu created by an inventory provider to allow the use of MiniMessage in the title.
- Added the `getButtonTemplateItem()` method to `AbstractMenuViewButton` so `AbstractPagedMenuButton` could get `TemplateItem`, and consequently `ItemBuilder`, instead of creating a new `ItemBuilder` from a pre-existing `ItemStack`. The previous approach prevented correct argument replacement; because if a line contained MiniMessage tags that changed color with every letter (like `<rainbow>`), the pre-existing `ItemStack` would already have converted them into color characters, making proper replacement impossible. Consequently, I also changed the abstract method in `AbstractPagedMenuButton` from `modifyViewItem(ItemStack)` to `modifyViewItem(ItemBuilder)` to allow direct access to the `ItemBuilder`.
- In `modifyViewItem()` within `TopIslandsSelfIslandButton`, instead of separately replacing the arguments for the name and then the lore, I combined the steps and perform them at the very end, after the member list has been correctly replaced from `{4}`.


OmerBenGera:
Great feature!
I don't like that there are 3 providers for MiniMessage now lol
I want to convert them into a single, one provider.
What do you think about this idea?

Lemvs:
> Great feature!
> 
> I don't like that there are 3 providers for MiniMessage now lol
> 
> I want to convert them into a single, one provider.
> 
> What do you think about this idea?

I also thought about it, just how to call it sensibly. MessagesProvider and InventoryProvider have a common goal because they just format text, so maybe TextProvider, but BossBarProvider also creates a BossBar implementation, so I don't know if it fits

OmerBenGera:
> > Great feature!
> > I don't like that there are 3 providers for MiniMessage now lol
> > I want to convert them into a single, one provider.
> > What do you think about this idea?
> 
> I also thought about it, just how to call it sensibly. MessagesProvider and InventoryProvider have a common goal because they just format text, so maybe TextProvider, but BossBarProvider also creates a BossBar implementation, so I don't know if it fits

Let's just merge InventoryProvider into MessagesProvider and keep BossBarPovider on its own - I am fine with that

src/main/java/com/bgsoftware/superiorskyblock/core/menu/button/AbstractPagedMenuButton.java
```diff
@@ -26,9 +27,22 @@ public E getPagedObject() {
 
     @Override
     public final ItemStack createViewItem() {
-        return modifyViewItem(super.createViewItem());
+        return modifyViewItem(getItemBuilder());
```
OmerBenGera:
```suggestion
        return modifyViewItem(Optional.ofNullable(getButtonTemplateItem()).map(TemplateItem::getBuilder).orElse(null));
```

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/external/inventory/InventoryProvider.java
OmerBenGera, line 1: Merge into `MessagesProvider`. Also, rename `MessagesProvider` into a "UIFactory" or something like that.

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/external/inventory/InventoryProvider.java
Lemvs: > Merge into `MessagesProvider`. Also, rename `MessagesProvider` into a "UIFactory" or something like that.

So if it's supposed to be a factory, should I move it to FactoriesManagerImpl? And how do I load the hook there? Or is that just the name?

OmerBenGera: Rename it to "UIProvider" and keep it as a provider in the providersmanager

OmerBenGera (approved):

PR #3003, opened 2026-08-02
Moved the configuration of features related to warps and visit to separate modules, and many more changes...

Lemvs:
First, I'd like to say that I know this is a lot of changes, but since I had to migrate everything anyway, I didn't want to edit it again. I tried to list most of the changes I made, but if I forgot something and you don't know why I changed something in the code, ask, and I'll explain everything.

# Really long changelog # 
- New "Visit" and "Warps" modules have been created, to which all configuration related to these features has been moved.
- The "global-warps" menu has been separated. Instead of displaying islands with warps or islands with visitor homes, it can now only display islands with warps, and islands with visitor homes will now be displayed in the new "visit-islands" menu. If the "global-warps" menu file was set to display visitor homeow, its appearance will be copied to the "visit-islands" menu, the "warps" command will be disabled, and its aliases will be added to the "visit" command to maintain consistency with the previous behavior. However, if you don't want to use the "visit-islands" menu, you can disable it in the module configuration.
- Changed the "warps" path to "slots" in the "global-warps" menu to maintain consistency with other menus and simplify configuration loading. If the "global-warps" menu file contains the old path, it will be converted accordingly.
- Changed "SwitchTopIslandsSortingTypeButton" to an abstract button, which is now also used by the "global-warps" and "visit-islands" menus. 
- Added a "dimension-selection" menu, whose purpose is to select a dimension and perform further steps specified as an argument to open that menu for the player. The menu itself doesn't specify the purpose of the selection; it's general, so it's currently used to select dimensions for the "biomes" and "visit-islands" menus. In the future, it will also be used to select dimensions for the "go" command or the "settings" menu.
- Removed "DimensionSelectionMode" because I think it's redundant after the introduction of the "dimension-selection" menu. In my opinion, this method of selecting dimensions is more intuitive. "DimensionSelectionMenu" wasn't yet in the stable API, so I think it can be safely removed.
- Disabled "SignsListener" registering if the "Visit" module is disabled and the "Warps" module is disabled or has signs disabled.
- Removed the warp limit check when placing a visitor sign, as it doesn't count towards the warp size.
- Added support for placing a visitor home outside the default dimension, which can be toggled in the "Visit" module configuration. Previously, it was possible to place a visitor sign outside the default dimension, even though the location associated with it was set in the default dimension, causing inconsistency.
- Added a separate message after placing a visitor home, instead of using the warp setting message.
- Added a message after destruction, deactivation, or teleportation to a visitor home.
- Added a message for island members about teleporting to their visitor home, provided they haven't vanished and are not a member of the island.
- Added the ability to change a visitor sign to inactive when someone tries to teleport to an unsafe location, not just when the old visitor sign is deactivated by placing a new one.
- Improved teleport location selection when visitor signs are disabled. Previously, even if an island had a visitor home, the island home was selected; now this will not be the case.
- Added the ability to disable the "warp-manage" and "warp-category-manage" menus in the "Warps" module configuration. These do not require any permissions and are not triggered by a separate command, so previously they could not be disabled.
- Added missing color code stripping for warp names set with the command or changed using "warp-manage menu" or "warp-category-menu." These names are used as identifiers in commands, so they should not contain color codes. If you want colorful names, you can edit the icon.
- Added the ability to change the minimum and maximum length of warp and warp category names, as well as blacklisted names that support regexes. Regex support has also been added for blacklisted island names. (https://github.com/BG-Software-LLC/SuperiorSkyblock2/discussions/1690)
- Changed the behavior of destroying the sign when deleting a warp or changing its location to updating its appearance when inactive. 
- Added an update to the warp sign's appearance after changing the warp name or warp category.
- Added the ability to display the warp category name on warp signs.
- Removed the setting of the warp location yaw and visitor home to the player's yaw when the island was created during the island blueprint insertion process, as the player didn't place the island, so it was pointless. **Furthermore, this option doesn't work at all (at least in version 26.2) because the square brackets are simply removed from the signs after the island is created (this was also the case before my changes). Could you tell me if this is how it should be?**
- Added the ability to translate the "-cancel" text when editing a warp or warp category.
- Added the ability to set warp privacy when creating a warp with a command. (https://github.com/BG-Software-LLC/SuperiorSkyblock2/discussions/1278)
- Added the ability to set warp category when creating a warp with a sign.
- Fixed warp names being tabbed in the "warp" command if the player name was used instead of the island name as an argument.
- Manual editing of warp and visitor home signs has been disabled; warps can be edited through the menu, but visitor home signs are not.
- Renamed many warp messages to match visitor home messages and start with the word warp so they are grouped together.
- Added the success update sound to play when toggling warp privacy in the menu. 
- Added the "Island#relocateWarp()" method, which should be used instead of IslandWarp#setLocation() (as is the case for rename), because without it, the map storing warps for locations in the island object is unaware of the location change, and therefore remembers outdated locations.
- Many other changes, both minor and major. If any require comment, please ask.

1. I added a lot of converts; most of them worked... until I removed unnecessary options from the main config. After that, the module configuration convert stopped working. This is because the main config is loaded first, and this [code](https://github.com/BG-Software-LLC/CommentedConfiguration/blob/bed3c4636915a8f808563f349a6d13fc62bfc646/src/main/java/com/bgsoftware/common/config/CommentedConfiguration.java#L419) causes paths that are in the server config but not in the default config to not be added after clearing, thus being lost, preventing modules from converting them later. In my opinion, "correctOrder" should be supplemented with keys that are in "target" but not in "section," because removing old options should be handled by whoever wrote convert, not the "syncWithConfig" method. Unless there's another way to load the module earlier, let me know what to do.
2. Visitor homes were created with the intention of being per-dimensional, but for reasons unknown to me, this was unused for many years. However, I don't understand why something that was supposed to be multiple was merged with the island description, which is only one. This now creates a problem – the description from which the visitor sign should be used. In my opinion, these options should be separated, because the descriptions are intended to describe the visitor home, not the island itself, and that's how they were used because they could only be found in the global-warps menu. I don't know how to solve this now, so tell me what you think. In my opinion, descriptions should be per-visitor home, but how do you do this sensibly?


OmerBenGera:
This PR will not be added to 2026.3, but to the next release (aka the dev builds after 2026.3)
Until then, I will do the CR but dont expect it to be approved until then :)

Lemvs:
> This PR will not be added to 2026.3, but to the next release (aka the dev builds after 2026.3)
> 
> Until then, I will do the CR but dont expect it to be approved until then :)

I understand, it's a lot of changes. But I deleted the DimensionSelectionMode enum due implementing MenuDimensionSelection, which was added a few dev versions ago, and if you release stable versions with it, you won't be able to just delete it, because it will be in the official API 😓

src/main/java/com/bgsoftware/superiorskyblock/core/menu/impl/MenuDimensionSelection.java
OmerBenGera, line 1: I don't see a reason for this menu tbh if we remove visitor locations from multiple dimensions.

src/main/java/com/bgsoftware/superiorskyblock/module/visit/utils/VisitUtils.java
```diff
@@ -0,0 +1,241 @@
...
+            Message.NO_CLOSE_BYPASS.send(superiorPlayer);
+            return;
+        }
+
+        if (isVisitorSign && !WorldBlocks.isSafeBlock(location.getBlock())) {
+            Message.VISITOR_HOME_TELEPORT_UNSAFE.send(superiorPlayer);
+
+            if (!superiorPlayer.hasBypassModeEnabled()) {
+                if (PluginEventsFactory.callIslandRemoveVisitorHomeEvent(island, superiorPlayer)) {
+                    island.setVisitorsLocation(null);
```
OmerBenGera, line 175: This is a bug - it always set the visitors location of the default world dimension.

src/main/java/com/bgsoftware/superiorskyblock/island/IslandNames.java
```diff
@@ -89,54 +94,90 @@ public static String getNameForLookup(String name) {
...
-        return !warpName.isEmpty() && warpName.length() <= getMaxWarpNameLength();
+        return !Text.isBlank(warpName) && warpName.length() >= BuiltinModules.WARPS.getConfiguration().getNamesMinLength()
+                && warpName.length() <= BuiltinModules.WARPS.getConfiguration().getNamesMaxLength();
     }
 
-    public static int getMaxWarpNameLength() {
-        return 255;
-    }
+    public static boolean isValidWarpName(@Nullable SuperiorPlayer superiorPlayer, Island island, @Nullable String warpName) {
+        if (Text.isBlank(warpName) || warpName.indexOf(" ") > 0) {
```
OmerBenGera, line 102:
```suggestion
        if (Text.isBlank(warpName) || warpName.indexOf(" ") >= 0) {
```

src/main/java/com/bgsoftware/superiorskyblock/module/warps/commands/CmdSetWarp.java
```diff
@@ -77,56 +81,73 @@ public Message getPermissionLackMessage() {
...
                 return;
+            }
         }
 
         WarpCategory warpCategory = categoryName == null ? null : island.createWarpCategory(categoryName);
 
+        boolean privateFlag = BuiltinModules.WARPS.getConfiguration().isPrivateByDefault();
+
+        if ((args.length == 4 && BuiltinModules.WARPS.getConfiguration().isCategoriesEnabled()) || args.length == 3) {
+            privateFlag = Boolean.parseBoolean(args[args.length - 1]);
```
OmerBenGera, line 123: The warp category will be parsed as the private flag like so:
```
/is setwarp XXX
```

src/main/java/com/bgsoftware/superiorskyblock/core/messages/Message.java
```diff
@@ -1015,17 +1049,35 @@ private static void convertOldFile() {
...
-            return true;
+            converted.set(true);
         }
+        convertMessageName(cfg, "INVALID_VISIT_LOCATION", "VISITOR_HOME_TELEPORT_NOT_SET", converted);
+        convertMessageName(cfg, "INVALID_VISIT_LOCATION_BYPASS", "VISITOR_HOME_TELEPORT_NOT_SET_BYPASS", converted);
+        convertMessageName(cfg, "DELETE_WARP", "WARP_DELETE", converted);
+        convertMessageName(cfg, "DELETE_WARP_SIGN_BROKE", "WARP_SIGN_DEACTIVATED", converted);
+        convertMessageName(cfg, "SET_WARP", "WARP_SET", converted);
+        convertMessageName(cfg, "SET_WARP_OUTSIDE", "WARP_SET_OUTSIDE_ISLAND", converted);
+        convertMessageName(cfg, "TELEPORTED_TO_WARP", "WARP_TELEPORT", converted);
```
OmerBenGera, line 1068: `WARP_TELEPORT` is not a valid enum

src/main/java/com/bgsoftware/superiorskyblock/module/visit/utils/VisitUtils.java
```diff
@@ -0,0 +1,241 @@
...
+            if (description.length() > 0) {
+                PluginEvent<PluginEventArgs.IslandChangeDescription> changeDescriptionEvent =
+                        PluginEventsFactory.callIslandChangeDescriptionEvent(island, superiorPlayer, description.toString());
+
+                if (!changeDescriptionEvent.isCancelled()) {
+                    island.setDescription(changeDescriptionEvent.getArgs().description);
+                }
+            }
+
+            if (BuiltinModules.VISIT.getConfiguration().isDescriptionsEnabled()) {
```
OmerBenGera, line 82: Always true

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
```diff
@@ -3721,73 +3749,91 @@ public void deleteCategory(WarpCategory warpCategory) {
...
+        Preconditions.checkNotNull(newLocation, "newLocation parameter cannot be null.");
+        if (!(newLocation instanceof LazyWorldLocation)) {
+            Preconditions.checkNotNull(newLocation.getWorld(), "location's world cannot be null.");
+        }
+        Preconditions.checkState(getWarp(newLocation) == null, "Cannot relocate warps to an already existing warp.");
+
+        try (ObjectsPools.Wrapper<LazyWorldLocation> wrapper = ObjectsPools.LAZY_LOCATION.obtain()) {
+            Location location = islandWarp.getLocation(wrapper.getHandle());
+            this.warpsByLocation.write(warpsByLocation -> warpsByLocation.remove(location));
+            this.warpsByLocation.write(warpsByLocation -> warpsByLocation.put(newLocation, islandWarp));
```
OmerBenGera, line 3818: Merge into the same write callback - not atomic

src/main/java/com/bgsoftware/superiorskyblock/commands/player/CmdBiome.java
```diff
@@ -72,54 +74,53 @@ public Message getPermissionLackMessage() {
...
+    private void closeMenu(SuperiorPlayer superiorPlayer) {
+        MenuView<?, ?> menuView = superiorPlayer.getOpenedView();
 
-    private boolean isMode(DimensionSelectionMode dimensionSelectionMode) {
-        return Menus.MENU_BIOMES.getDimensionSelectionMode() == dimensionSelectionMode;
+        if (menuView != null && menuView.getMenu() instanceof MenuDimensionSelection) {
+            menuView.setPreviousMove(false);
+            BukkitExecutor.sync(menuView::closeView, 1L);
+        }
     }
```
OmerBenGera, line 124: You had to do this because the DimensionSelectionMenu is not dynamic.
You must remove this and any other places where you did this hack.

src/main/java/com/bgsoftware/superiorskyblock/core/menu/impl/MenuVisitIslands.java
OmerBenGera, line 1: I don't see why would you want to support multiple dimensions for visitor locations.
There's a reason it was never implemented - you only need one visitor location. The better feature would be visitors location supported in any dimension, but yet again - only one at a time.

src/main/java/com/bgsoftware/superiorskyblock/module/visit/VisitModule.java
```diff
@@ -0,0 +1,218 @@
...
+import java.util.Locale;
+
+public class VisitModule extends BuiltinModule<VisitModule.Configuration> {
+
+    public VisitModule() {
+        super("visit");
+    }
+
+    @Override
+    protected boolean onConfigCreate(SuperiorSkyblockPlugin plugin, CommentedConfiguration moduleConfig, boolean firstTime) {
```
OmerBenGera, line 25: When reading the main config in this time, it was already been called with `syncWithConfig`. This causes the sections you try to access to not be in the config file.
This is why `convertData` in SettingsManagerImpl is called before `syncWithConfig`, and it should be responsible for converting the data into the module's config files - not the opposite.

src/main/java/com/bgsoftware/superiorskyblock/island/SIsland.java
```diff
@@ -283,6 +282,7 @@ protected IslandCache create() {
     private volatile String formattedName;
     private volatile String strippedName;
     private volatile String description;
+    private final Synchronized<EnumerateMap<Dimension, String>> descriptions = Synchronized.of(new EnumerateMap<>(Dimension.values()));
```
OmerBenGera, line 285: Remove

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/module/visit/VisitModule.java
Lemvs: Then how did it work in the bank module, it hasn't been changed for a long time, and it's done the same way - in BankModule, not in SettingsManagerImpl https://github.com/BG-Software-LLC/SuperiorSkyblock2/blob/dev/src/main/java/com/bgsoftware/superiorskyblock/module/bank/BankModule.java#L29

src/main/java/com/bgsoftware/superiorskyblock/core/menu/impl/MenuDimensionSelection.java
Lemvs: For Biomes Menu, in the future maybe also for Settings Menu or /is go, as I wrote in the Pull Request description

src/main/java/com/bgsoftware/superiorskyblock/module/visit/VisitModule.java
OmerBenGera: > Then how did it work in the bank module, it hasn't been changed for a long time, and it's done the same way - in BankModule, not in SettingsManagerImpl https://github.com/BG-Software-LLC/SuperiorSkyblock2/blob/dev/src/main/java/com/bgsoftware/superiorskyblock/module/bank/BankModule.java#L29

Probably didnt work and never been reported

src/main/java/com/bgsoftware/superiorskyblock/core/menu/impl/MenuVisitIslands.java
Lemvs: So what was the point of all those per-dimension methods, or even storing those locations in a map, if there’s only supposed to be one of them? If there’s only supposed to be a single location, but in any dimension, then I probably shouldn't even store it as a map, after all, how would I go about collecting those locations later? (Unless I just took the first element of the map, but that doesn't really make sense.)

src/main/java/com/bgsoftware/superiorskyblock/module/visit/utils/VisitUtils.java
Lemvs: What? That value is fetched from the config, so why would it always be true? I don't understand.

PR #3015, opened 2026-08-20
Added support for SmartSpawner

Lemvs:
As suggested in the discussion https://github.com/BG-Software-LLC/SuperiorSkyblock2/discussions/3014

This plugin is quite... specific, which meant that:
- I had to disable `listenToSpawnerChanges` in `ProvidersManagerImpl` for this hook because, while the standard `BlockPlaceEvent` fires normally in Creative mode, it gets cancelled in Survival mode, leaving me with only their `SpawnerPlaceEvent` to work with.
- I had to add a `listenToSpawnerChangesUsingEggs` option, because our `UpgradeTypeBlockLimit` detects the change and sends a message, but the spawner type is still changed, so I need to listen for their `SpawnerEggChangeEvent` to cancel the action appropriately.


Lemvs:
We have to wait a moment, because this needs to be resolved: https://github.com/OpenVdra/SmartSpawner/issues/311#issuecomment-5379555109

Lemvs:
I simply set the condition the other way around, but even so, if they resolve it, their API can be updated to avoid that issue with spawn eggs, which they’ve actually fixed now.

Lemvs:
Okay, everything should be working properly now.

OmerBenGera (approved):

PR #3016, opened 2026-08-21
Added support for Nova

Lemvs:
As suggested in discussion https://github.com/BG-Software-LLC/SuperiorSkyblock2/discussions/1922

Unfortunately, this required changing the way keys are created. Previously, if a key contained more than one colon, everything following the second colon was lost. Now, the part before the first colon is treated as the global-key, and the rest as the sub-key. 

This change stems from the fact that Nova identifies items using namespaced keys, consequently, multiple items can share the same name but belong to different namespaces. I couldn't use the approach from the CraftEngine hook (creating a list of available items where colons in the namespaced key were replaced with underscores) because Nova doesn't expose a list of all items or namespaces. Theoretically, I could retrieve a list of items sharing a specific name, but since there might be several, I wouldn't know which one is the correct one. Attempting to iteratively replace underscores with colons to see if an item is found wouldn't make much sense either, which is why I made this decision.


OmerBenGera (changes requested):
The plugin doesn't really work for me - did you test it out? It works well?

Lemvs:
> The plugin doesn't really work for me - did you test it out? It works well?

I had trouble with it too, but I eventually got it working, and I double-checked everything I did.

I usually run my server on a free host, but this plugin requires changing the startup parameters, which I can't do on that host, so I run my server on localhost. Did you definitely add `javaagent` to the startup parameters? https://docs.xenondevs.xyz/nova/admin/setup/

Additionaly, I tested this on Paper 26.2 and the latest version of Nova (0.24.0).

And in the config, I set something like this so the texture pack would load automatically:
```
# Configuration options for the resource pack
# READ SETUP GUIDE: https://xenondevs.xyz/docs/nova/admin/setup/#step-2-resourcepack-hosting
resource_pack:
  # The auto uploader automatically uploads the resource pack after generation.
  # READ SETUP GUIDE: https://xenondevs.xyz/docs/nova/admin/setup/#automatic-resource-pack-hosting
  auto_upload:
    enabled: true
    service: self_host
    port: 38519
    host: 127.0.0.1
    append_port: true
```

Then I added the Machines add-on (normally placed in the plugins folder) along with the Simple Upgrades plugin, which also goes there, because by default Nova only has one item, and I needed blocks for tests. https://modrinth.com/plugin/nova-machines

OmerBenGera (approved):

PR #3017, opened 2026-08-22
Added support for MythicMobs

Lemvs:
As suggested in the discussions https://github.com/BG-Software-LLC/SuperiorSkyblock2/discussions/1524 https://github.com/BG-Software-LLC/SuperiorSkyblock2/discussions/150

# Changelog #
- Although CustomKeyParser was designed to work with both blocks and entities, its previous registration method prevented the use of entities; this was because BlockValuesManagerImpl created a KeyMap using the material's KeyIndicator, making it impossible to include entities. Consequently, a separate method was required, so I took the opportunity to move this functionality from BlocksValuesManager to KeysManager (as it is no longer strictly tied to block values, being used for limits, missions etc. as well), along with other helper methods like convertKey(...).
- I deprecated the recordEntitySpawn(EntityType, Location) and recordEntityDespawn(EntityType, Location) methods in WorldRecordService. In their place, I created corresponding methods that replace EntityType with Key (matching the approach used for recordBlock...()), as the old methods prevented the correct detection of custom entities, since they cannot be identified based on EntityType alone.


API/src/main/java/com/bgsoftware/superiorskyblock/api/service/world/WorldRecordService.java
```diff
@@ -100,9 +100,21 @@ RecordResult recordBlockPlace(Key blockKey, Location blockLocation, int blockCou
      * @param entityType The type of the entity that spawned.
      * @param location   The location of where the entity was spawned.
      * @return The result of recording the entity spawn.
+     * @deprecated See {@link WorldRecordService#recordEntitySpawn(Key, Location)}
+     *             Do not use the old method, as it will not detect custom mobs.
      */
+    @Deprecated
```
OmerBenGera, line 106: Don't deprecate it, keep it.
Similar to how `recordBlockPlace` can take `Block`

Hooks/MythicMobs/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+    mavenCentral()
+    maven { url 'https://mvn.lumine.io/repository/maven-public/' }
+}
+
+dependencies {
+    compileOnly 'io.lumine:Mythic-Dist:5.13.0'
+    compileOnly 'org.spigotmc:v1_8_R3-Taco:latest'
+    compileOnly project(":API")
+    compileOnly project(":")
+}
```
OmerBenGera:
```suggestion
dependencies {
    compileOnly 'io.lumine:MythicMobs:5.0.1'
    compileOnly 'org.spigotmc:v1_8_R3-Taco:latest'
    compileOnly project(":API")
    compileOnly project(":")
}
```

src/main/java/com/bgsoftware/superiorskyblock/core/key/KeysManagerImpl.java
```diff
@@ -27,6 +31,9 @@
 
 public class KeysManagerImpl extends Manager implements KeysManager {
 
+    private static final KeyMap<CustomKeyParser> customMaterialKeyParsers = KeyMaps.createArrayMap(KeyIndicator.MATERIAL);
+    private static final KeyMap<CustomKeyParser> customEntityTypeKeyParsers = KeyMaps.createArrayMap(KeyIndicator.ENTITY_TYPE);
+
```
OmerBenGera: These lists can be large, let's convert them into a HashMap based key-maps.

src/main/java/com/bgsoftware/superiorskyblock/service/world/WorldRecordServiceImpl.java
```diff
@@ -256,38 +256,59 @@ private void recordBlockBreakInternal(Island island, Key blockKey, Location bloc
...
+        }
 
-        if (BukkitEntities.canBypassEntityLimit(entity))
+        if (BukkitEntities.canBypassEntityLimit(entity)) {
             return RecordResult.ENTITY_CANNOT_BE_TRACKED;
+        }
+
+        if (!isEntityLimitsEnabled() || !BukkitEntities.canHaveLimit(entity.getType())) {
+            return RecordResult.ENTITY_CANNOT_BE_TRACKED;
+        }
```
OmerBenGera, line 269: `isEntityLimitsEnabled()` should be the first check

OmerBenGera (changes requested):
Great feature :)

API/src/main/java/com/bgsoftware/superiorskyblock/api/service/world/WorldRecordService.java
Lemvs: But Block is a completely different situation, because it isn't just an enum element but an actual object in the world; this allows us to use a CustomKeyParser to retrieve a custom key for it. The comparison would be valid if there were a method accepting a Material but not a Block - Block is actually a good approach, whereas EntityType definitely isn't. If we provide an EntityType, we will skip the CustomKey check, so, such a method shouldn't exist either, because it could introduce an inconsistency between what is currently on the island and how it was counted.

src/main/java/com/bgsoftware/superiorskyblock/service/world/WorldRecordServiceImpl.java
Lemvs: But the condition you commented on already follows that order, unless you mean the order of conditions within the entire method? Please clarify

PR #3018, opened 2026-08-22
Added support for mcMMO (auto replant)

### LINKED ISSUE #2383: Farming Mission don't work with McMMO skill
### Minecraft's Version

AdvancedSlimePaper version 1.21.3-16525-dev/1.21.3@d5ade1e (2024-12-15T21:45:37Z) (Implementing API version 1.21.3-R0.1-SNAPSHOT)

### Plugin's Version

SuperiorSkyblock2 version 2024.4-b430

### Describe the bug

Island farming missions don't work with McMmo's auto replant. It doesn't count the crop, and it's a real handicap when players have an 80% chance of replanting automatically...

### To Reproduce

1. Install McMmo
2. Try to plant a crop with autoreplant

### Additional Information

_No response_

Lemvs:
Fixes #2383 

This probably isn't the best way to solve the problem, but it's hard to come up with something better. I don't have access to `FarmingMissions` or `PlantTracker` here, and creating a custom API event, calling it in a hook and listening for it in the mission, seems like a poor approach too, so, reflection seems like the best option, unless you have a better idea?


OmerBenGera (changes requested):
The best place to put this hook is in `FarmingMissions`, not SSB
If that mission was maintained by a third-party developer, he would have to add support to mcmmo. Therefore, its the mission's implementation responsibility to support it - therefore, it should go there

Lemvs:
> The best place to put this hook is in `FarmingMissions`, not SSB If that mission was maintained by a third-party developer, he would have to add support to mcmmo. Therefore, its the mission's implementation responsibility to support it - therefore, it should go there

Okay, but how exactly can I do that?

PR #3020, opened 2026-08-23
Added entity category limits

Lemvs:
As suggested in the discussions https://github.com/BG-Software-LLC/SuperiorSkyblock2/discussions/1134 https://github.com/BG-Software-LLC/SuperiorSkyblock2/discussions/1866 (and for blocks https://github.com/BG-Software-LLC/SuperiorSkyblock2/discussions/164)

It would be nice if we could finally add block categories (#2927) as well, so I could do the same for them, so for now, this Pull Request can wait until then to implement it simultaneously (just hope it doesn't wait as long as the one with block categories.....).

# Changelog #
- Fixed the creation of default entity categories; they were previously created twice due to the difference in capitalization of their names.
- Added a new table, `islands_entity_category_limits`, which stores entity category limits for islands.
- Added the `default-values.entity-category-limits` option in the main config, which sets the default entity category limits; it's empty by default.
- Added the option to add `entity-category-limits` in upgrade levels, which sets the entity category limits for a given upgrade level. 
- Added the `/island admin addentitycategorylimit`, `/island admin removeentitycategorylimit`, and `/island admin setentitycategorylimit` commands, which allow you to manage entity category limits for an island.
- Added the `superior_island_entity_category_count_<#>` and `superior_island_entity_category_limit_<#>` placeholders, which allow you to check the number of entities in a given entity category and their limit.
- Added the display of entity category limits for an island in `/island admin show`.
- Added new upgrade methods, corresponding to existing methods related to entity limits, but for entity category limits. The storage of entity category limits is based on how limit roles are stored, because, like them, they are recreated after a plugin reload, so they objects cannot be stored directly in limit maps, I store their names as keys in maps.
- Added new methods for the island and its builder, corresponding to existing methods related to entity limits, but for entity category limits. The storage of entity limit categories is based on how limit roles are stored. Like limit roles, they are recreated upon plugin reload, so they objects cannot be stored directly in limit maps, I store their names as keys in maps.
- The principle of the island entity tracker remains the same; there is simply an additional map that stores the number of mobs for a given category. Because entity categories are dynamic, this map is cleared upon plugin reload and repopulated.


nivcoo:
Good feature !

Lemvs:
> Good feature !

Thanks 😢

PR #3029, opened 2026-09-01
Improved and fixed some commands

Lemvs:
# Changelog #
- Added support for multiple islands to the `/is admin close` command; added distinct messages depending on whether a player name, island name, or `*` was entered.
- Added support for multiple islands to the `/is admin open` command; added distinct messages depending on whether a player name, island name, or `*` was entered.
- Added support for multiple islands to the `/is admin ignore` command; added distinct messages depending on whether a player name, island name, or `*` was entered; removed already ignored islands from tab completion.
- Added support for multiple islands to the `/is admin unignore` command; added distinct messages depending on whether a player name, island name, or `*` was entered; removed already unignored islands from tab completion.
- Added a response for the admin after using `/is admin kick`.
- Added an `ALL` message variant for the `/is admin deposit`, `/is admin cmdall`, `/is admin msgall`, and `/is admin titleall` commands, replacing the practice of using the island name variant with 'all' inserted in that spot.
- Secured the `/is admin deposit` command against depositing negative amounts; the `/is admin withdraw` command already had this safeguard, so it was necessary here as well to prevent the bank balance from going negative.
- Added the ability to translate the "online-filter" argument name in the `/is admin cmdall` command, similar to the "save-air" argument in the `/is admin schematic` command.
- Added the display of the "load/unload" tab complete in the `/is admin modules` command.
- Added a separate message for the `/is admin removeratings` command when an island name is provided, so it does not use the same message as when a player name is provided.
- Removed placeholder parsing within the method SIsland#sendMessage(), since Message.CUSTOM.send() creates the message component, and sending it requires retrieving the message content (which also parses placeholders), there is no need to perform this step twice.
- Fixed the display of schematic names in tab completions and their usage in arguments; since custom dimensions can now be created, simply checking if a schematic name ends with "normal", "nether", or "the_end" is no longer sufficient, as other dimensions may exist.


src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminKick.java
```diff
@@ -71,6 +71,7 @@ public void execute(SuperiorSkyblockPlugin plugin, CommandSender sender, Superio
 
         IslandUtils.handleKickPlayer(sender instanceof Player ? plugin.getPlayers().getSuperiorPlayer(sender) : null,
                 sender.getName(), targetIsland, targetPlayer);
+        Message.KICKED_MEMBER.send(sender, targetPlayer.getName());
```
OmerBenGera: `handleKickPlayer` can fail, but `KICKED_MEMBER` sends regardless.

src/main/java/com/bgsoftware/superiorskyblock/commands/admin/CmdAdminClose.java
OmerBenGera: Add `COMMAND_ARGUMENT_ALL_ISLANDS`. Also, do this for all the other commands

OmerBenGera (changes requested):

PR #3030, opened 2026-09-01
refactor: performance improvements

reecepbcups:
### Improvements

1. every island lookup (block break/place, redstone, spawn, interact, movement) is 3.78x faster and reduces 40bytes per lookup
2. De duplicated getIslandAt(from) on move/teleport **(biggest CPU performance win)**
3. Safe spot optimization that uses 256x less allocations (from 10,240 to just 40B.op)
4. areAllOnlinePlayersInsideAFK is 2.6x faster whent his setting is enabled (157 -> 3B/call) **(biggest RAM performance win)**
5. dateFormatter ThreadLocal fix

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/island/container/DefaultIslandsContainer.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/island/container/DefaultIslandsContainer.java
@@ -11,6 +11,7 @@ import com.bgsoftware.superiorskyblock.api.wrappers.BlockPosition;
 import com.bgsoftware.superiorskyblock.core.IslandPosition;
 import com.bgsoftware.superiorskyblock.core.LazyWorldLocation;
 import com.bgsoftware.superiorskyblock.core.SWorldPosition;
+import com.bgsoftware.superiorskyblock.island.SIsland;
 import com.bgsoftware.superiorskyblock.core.SequentialListBuilder;
 import com.bgsoftware.superiorskyblock.core.collections.EnumerateSet;
 import com.bgsoftware.superiorskyblock.core.events.plugin.PluginEventType;
@@ -158,7 +159,16 @@ public class DefaultIslandsContainer implements IslandsContainer {
     public Island getIslandAt(Location location) {
         Island island = plugin.getProviders().hasCustomWorldsSupport() ?
                 customWorldsSupportIslandLookup(location) : nativeIslandLookup(location);
-        return island == null || !island.isInside(SWorldPosition.of(location)) ? null : island;
+        if (island == null)
+            return null;
+
+        // Fast path: verify the X/Z area intercept without allocating a WorldPosition per lookup.
+        // Semantics match isInside(WorldPosition) (area-only, no world check). Non-SIsland islands
+        // fall back to the original path to preserve behavior.
+        boolean inside = island instanceof SIsland ?
+                ((SIsland) island).intersectsArea(location.getX(), location.getZ()) :
+                island.isInside(SWorldPosition.of(location));
+        return inside ? island : null;
     }
 
     private Island customWorldsSupportIslandLookup(Location location) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/bgsoftware/superiorskyblock/island/container/DefaultIslandsContainer.java
```diff
@@ -158,7 +159,16 @@ public int getIslandsAmount() {
...
-        return island == null || !island.isInside(SWorldPosition.of(location)) ? null : island;
+        if (island == null)
+            return null;
+
+        // Fast path: verify the X/Z area intercept without allocating a WorldPosition per lookup.
+        // Semantics match isInside(WorldPosition) (area-only, no world check). Non-SIsland islands
+        // fall back to the original path to preserve behavior.
+        boolean inside = island instanceof SIsland ?
+                ((SIsland) island).intersectsArea(location.getX(), location.getZ()) :
+                island.isInside(SWorldPosition.of(location));
```
OmerBenGera: You should never access SIsland directly.
If you want to get rid of this allocation, a better approach is to use ObjectPools. Create a new object called "MutableWorldPosition" which implements WorldPosition and ObjectsPool.Releasable.

src/main/java/com/bgsoftware/superiorskyblock/listener/IslandWorldEventsListener.java
```diff
@@ -52,7 +52,7 @@ private void onBlockRedstone(GameEvent<GameEventArgs.BlockRedstoneEvent> e) {
 
         if ((plugin.getSettings().isDisableRedstoneOffline() && !island.isCurrentlyActive()) ||
                 (plugin.getSettings().getAFKIntegrations().isDisableRedstone() &&
-                        island.getAllPlayersInside().stream().allMatch(SuperiorPlayer::isAFK))) {
+                        ((SIsland) island).areAllOnlinePlayersInsideAFK())) {
```
OmerBenGera: Same as before, never access SIsland directly.

src/main/java/com/bgsoftware/superiorskyblock/listener/IslandWorldEventsListener.java
```diff
@@ -76,7 +76,7 @@ private void onEntitySpawn(GameEvent<GameEventArgs.EntitySpawnEvent> e) {
             island = plugin.getGrid().getIslandAt(entity.getLocation(wrapper.getHandle()));
         }
 
-        if (island == null || island.isSpawn() || !island.getAllPlayersInside().stream().allMatch(SuperiorPlayer::isAFK))
+        if (island == null || island.isSpawn() || !((SIsland) island).areAllOnlinePlayersInsideAFK())
```
OmerBenGera: Same as before, never access SIsland directly.

OmerBenGera (changes requested):

src/main/java/com/bgsoftware/superiorskyblock/island/container/DefaultIslandsContainer.java
reecepbcups: Resolved with 06cc63bb78a662c8f24731a2386d92a0f23e656b

API/src/main/java/com/bgsoftware/superiorskyblock/api/island/Island.java
```diff
@@ -126,6 +126,15 @@ public interface Island extends Comparable<Island>, IMissionsHolder, IPersistent
...
     List<SuperiorPlayer> getAllPlayersInside();
 
+    /**
+     * Check whether all the online players that are on the island are AFK.
+     * An island with no online players inside is considered AFK.
+     * <p>
+     * This is an allocation-free equivalent of iterating {@link #getAllPlayersInside()}, intended
+     * for hot event paths.
+     */
+    boolean areAllOnlinePlayersInsideAFK();
```
OmerBenGera: Let's add a function that returns true/false for the playersInside list given a Predicate as an argument.
This way the exposed function is much more generic and can be used for other checks on the playersInside list.
We can expand this for all types of lists the island is tracking (members, visitors, etc)
Also, let's add a count method to simply return the count without creating a list.

OmerBenGera (changes requested):

API/src/main/java/com/bgsoftware/superiorskyblock/api/island/Island.java
reecepbcups: Resolved 2bcd1c1376f57a4d04e2cb789cb7cfe858814834

PR #3035, opened 2026-09-09
improv. rename TAMEABLE to TAMED and update entity category handling

nivcoo:
Hello,

Untamed animals, including llamas wolf etc, were incorrectly subject to TAMED_ANIMAL_DAMAGE because the TAMEABLE category matched every animal capable of being tamed, regardless of its actual state.

This PR applies that category only to animals that are already tamed and renames it to TAMED to clarify its meaning. Existing TAMEABLE configuration sections are automatically migrated, preserving their settings.

### DIFF
--- a/src/main/java/com/bgsoftware/superiorskyblock/config/SettingsContainer.java
+++ b/src/main/java/com/bgsoftware/superiorskyblock/config/SettingsContainer.java
@@ -747,7 +747,26 @@ public class SettingsContainer {
             removeInvalidEntityKeys = true;
         }
 
-        YamlConfiguration cfg = YamlConfiguration.loadConfiguration(file);
+        CommentedConfiguration cfg = CommentedConfiguration.loadConfiguration(file);
+
+        ConfigurationSection tameable = cfg.getConfigurationSection("TAMEABLE");
+        if (tameable != null) {
+            if (!cfg.contains("TAMED"))
+                cfg.createSection("TAMED");
+
+            for (Map.Entry<String, Object> entry : tameable.getValues(true).entrySet()) {
+                String path = "TAMED." + entry.getKey();
+                if (!(entry.getValue() instanceof ConfigurationSection) && !cfg.contains(path))
+                    cfg.set(path, entry.getValue());
+            }
+
+            cfg.set("TAMEABLE", null);
+            try {
+                cfg.save(file);
+            } catch (IOException error) {
+                Log.errorFromFile(error, file.getName(), "Failed to save converted entity categories:");
+            }
+        }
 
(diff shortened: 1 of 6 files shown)


nivcoo:
Updated the implementation to keep both categories:
- TAMEABLE applies to untamed animals that can be tamed.
- TAMED applies only to animals that are already tamed.

They are now mutually exclusive.
The migration no longer renames the entire TAMEABLE section. It only moves the legacy DAMAGE: TAMED_ANIMAL_DAMAGE mapping when TAMED does not already exist, preserving other settings and existing category definition

OmerBenGera (changes requested):
I am pretty sure TAMEABLE is only applied to acutal tamed animals.
This is done by calling BukkitEntities#isTameable on the entity in the `BukkitEntities#getCategories` function.
I don't think splitting these into two different categories for the default config makes sense. `TAMEABLE` for only tameable entities can be created manually, and `TAMED` is already the old `TAMEABLE` category.
Therefore, let's stick with only one category called `TAMEABLE` (for backwards compatibility), and fix the bug that caused entities that are not actually tamed to be considered in the `TAMEABLE` category.

nivcoo:
No if I do firstly the PR it’s because tameable is added even if the entity is not tamed. And after I discuss on another PR with @Lemvs and I do another commit to add BOTH like that if we want to target only tamed or only tameable we can

nivcoo:
With my first commit I convert tameable to tamed to convert old format to new and with second I convert only the old method too

OmerBenGera:
> No if I do firstly the PR it’s because tameable is added even if the entity is not tamed. And after I discuss on another PR with @Lemvs and I do another commit to add BOTH like that if we want to target only tamed or only tameable we can

`TAMEABLE` category should only be applied to tamed entities

nivcoo:
But the name TAMEABLE is not accurate for tamed entities that’s why my first commit convert it to TAMED and apply only the tag to tamed entities. So what do you want both tags or just TAMED ? Because keep tameable for tamed entities is not very good for the future

nivcoo:
Because depends on what do you want I can just revert my last commit and my first one fix the tamed check and convert the name correctly for old configs

OmerBenGera:
> But the name TAMEABLE is not accurate for tamed entities that’s why my first commit convert it to TAMED and apply only the tag to tamed entities. So what do you want both tags or just TAMED ? Because keep tameable for tamed entities is not very good for the future

Only `TAMED`

nivcoo:
It's good, now TAMEABLE is converted to TAMED and the "TAMED" category is added only when the entity is TAMED

PR #3036, opened 2026-09-10
improv. add support for Folia in NMS implementations and update task scheduling

nivcoo:
Hello ! 

I added the real support off folia. To be honest i used maybe for 60% of this work Astra (gpt 6), because the plugins has lot of scheduler and files...

The world API isn't available yet but i used NMS likle the folia Plugin WORLD, I had 2 choice, use World's plugin API or implement NMS directly.

I tested and it works pretty well for all features (i removed 0 feature).


Happy-FZM:
I used your method to add Folia support for 26.1, but when creating an island, the island’s border does not display for players after they are teleported to the island. Does 26.2 also have the same issue?
image

